PR #2, opened 2018-10-22
Add cost switch to recipes command

Tanookami:
Adds the /cost switch to the Recipes command. This returns a message about the cost of a recipe the user has not learned yet. There are a few trims for efficiency to other messages sent to users by this command.
Though the help file lists it, the /cost switch functionality actually didn't exist yet.
Resolves Arx in-game bug ticket #12057

### DIFF
--- a/commands/commands/crafting.py
+++ b/commands/commands/crafting.py
@@ -795,16 +795,16 @@ class CmdRecipes(ArxCommand):
 
     def display_recipes(self, recipes):
         from server.utils import arx_more
+        if not recipes:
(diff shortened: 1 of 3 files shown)


TehomCD (changes requested):
This looks really good to me, but maybe we should add a quick test for recipe cost - the crafting module doesn't have any test coverage at all.

TehomCD (approved):

TehomCD:
Awesome! Merged with a squeesh.

PR #3, opened 2018-10-22
Add a step in the re-roster process to clear proteges

Tanookami:
Adds a clear() for django to break the association between a rostered patron and any proteges they might have had.
While the re-rostering process cleared the relationship with a patron, it does not clear relationships with proteges. Adding this step makes the roster reset process cleaner for the next player.
This resolves arx in-game ticket #12206


TehomCD:
Looks good! Merging.

PR #4, opened 2018-10-22
Change iamhelping to consult actual action point cap

Tanookami:
The sum of the target's current action points and caller's proposed donation is compared to max_action_points property (as defined in RosterEntry,) instead of an arbitrary number.
Allows more flexibility for the AP cap to be adjusted without considering its impact on other commands.
This also resolves Arx in-game ticket #12060


TehomCD (changes requested):
Looks good, but I'd like to add a test for `+iamhelping` to `SocialTests` for this.

TehomCD (approved):

PR #6, opened 2018-10-26
This commit does the following:

tellg:

* Add the concept of the "displayable" idict property for containers. This allows
the contents of containers to be observed whether or not they are locked.

* Adds a 'root' and 'unroot' command that can only be targeted at container (and
not wearable container) objects. If a user is a decorator, it modifies the 'get'
lock to restrict access to accounts with the Builder perm, or characters with
decorator access on the room.

* Provides the ability to allow objects to be shown off (trophy cases, etc), without making them easy to steal. 

* Also, allows for the rooting of containers, like chests, etc.


TehomCD:
Looks good! Merged with rebase.

PR #8, opened 2018-10-31
This commit provides the basis for the template system to be used in …

tellg:

This adds the basis for a managed template system.

Reduce DB bloat and prevent ASCII theft.


world/templates/models.py
```diff
@@ -0,0 +1,82 @@
...
+    attribution = models.CharField(max_length=60)
+    apply_attribution = models.BooleanField(default=False)
+
+    title = models.CharField(max_length=255)
+
+    applied_to = models.ManyToManyField('objects.ObjectDB', blank=True)
+
+    grantees = models.ManyToManyField('character.RosterEntry', through="TemplateGrantee", blank=True)
+
+    accessible = TemplateManager()
```
TehomCD: Not a big deal, but I prefer managers that add methods rather than changing the default queryset replace 'objects' so that they're used by default and I don't need to remember the manager name as well as their methods.
```suggestion
    objects = TemplateManager()
```

world/templates/migrations/0001_initial.py
```diff
@@ -0,0 +1,59 @@
...
+                ('attribution', models.CharField(max_length=60)),
+                ('apply_attribution', models.BooleanField(default=False)),
+                ('title', models.CharField(max_length=255)),
+                ('applied_to', models.ManyToManyField(blank=True, to='objects.ObjectDB')),
+            ],
+            options={
+                'abstract': False,
+            },
+            managers=[
+                ('accessible', django.db.models.manager.Manager()),
```
TehomCD, line 35: Changing 'accessible' to 'objects' would change the migration here, I think. I can't recall ever seeing the managers kwarg in CreateModel before, so not entirely sure if it'd just be 'objects', or removed as a default, or what.

TehomCD (changes requested):
Looking great so far! Paxforms are still pretty new to me so I think I might look a bit more at the command, form, and tests tomorrow, but tentatively everything looks good. I made a small suggestion to change the name of the manager, since I prefer to replace the default if a custom manager just adds methods and doesn't change the default queryset, and I think that might need tweaking the migration. I like how it's looking, though!

emonical:
Also, I added the necessary work to get 'look' working. Still need to add the part for locking it onto descs and mapping the objects that have templates linked, but, getting there.

world/templates/models.py
emonical: Resolved in new push.

world/templates/migrations/0001_initial.py
emonical: Resolved with new migration (from push).

typeclasses/mixins.py
```diff
@@ -509,6 +510,9 @@ def return_appearance(self, pobject, detailed=False, format_desc=False,
                 string += "\n%s{n" % desc
         else:  # for crafted objects, respect formatting
             string += "\n%s{n" % desc
+            templates = Template.objects.from_desc(string)
```
TehomCD: What we probably wanna do here is have some simple caching with a `self.ndb` value. Like:
```
if self.ndb.cached_string_after_template_replacement:
    string = self.ndb.cached_string_after_template_replacement
else:
    templates = Template.objects.from_desc(string)
    if templates.exists():
        string = replace_template_values(string, templates)
    self.ndb.cached_string_after_template_replacement = string
```
then have both Template's save method and the desc property setter wipe set that value to `None`.

emonical: Ah, yeah, easy to tweak.

emonical: Resolved.

emonical: Oh, hmm, crap. The caching works, but, we're gonna have to wait until I have the template object relationship finished as well, for the cascade from Template save.

emonical:
Updated. Only piece missing (I think) is the desc setter uncaching, which, I'll admit to being unsure of.

TehomCD:
What you probably wanna do there is in the [three different desc property setters](https://github.com/Arx-Game/arxcode/blob/stable_orphan/typeclasses/mixins.py#L41), you'd add `self.ndb.cached_string_after_template_replacement = None` at the end. Each of those setters should be called whenever an object's description is changed in some way, either permanently or temporarily (like if they wore a mask that had a template, for example).

I wouldn't really worry about invalidating the cache if the template itself changes, thinking about it - that'll be done by staff I think if we're not giving players the ability to change them, so they can invalidate it manually if they hafta, or reload the server or whatever.

typeclasses/mixins.py
```diff
@@ -66,6 +68,7 @@ def __temp_desc_set(self, val):
             self.db.raw_desc = self.db.desc
         if not self.db.general_desc:
             self.db.desc = self.db.desc
+        self.ndb.cached_string_after_template_replacement = None
```
TehomCD: Whoops, I think I suggested the wrong variable name, doesn't match the attribute below. Should be:

```suggestion
        self.ndb.cached_template_desc = None
```

typeclasses/mixins.py
```diff
@@ -78,6 +81,7 @@ def __temp_desc_del(self):
         if not self.db.general_desc:
             self.db.general_desc = self.db.desc
         self.db.desc = ""
+        self.ndb.cached_string_after_template_replacement = None
```
TehomCD:
```suggestion
        self.ndb.cached_template_desc = None
```

typeclasses/mixins.py
```diff
@@ -93,6 +97,7 @@ def __perm_desc_set(self, val):
         """
         self.db.general_desc = val
         self.db.raw_desc = val
+        self.ndb.cached_string_after_template_replacement = None
```
TehomCD:
```suggestion
        self.ndb.cached_template_desc = None
```

emonical: done

typeclasses/mixins.py
emonical: done

typeclasses/mixins.py
emonical: done

typeclasses/mixins.py
emonical: This has been resolved.

world/templates/migrations/0002_auto_20181031_1455.py
```diff
@@ -0,0 +1,20 @@
...
+        ('templates', '0001_initial'),
+    ]
+
+    operations = [
+        migrations.AlterModelManagers(
+            name='template',
+            managers=[
+            ],
+        ),
+    ]
```
TehomCD, line 20: So the only other thing I'm thinking about is it might be a good idea to do a data migration to convert existing books into the template system. If you haven't written data migrations in django before, you'd define a function that has the arguments of `(apps, schema_editor)` for `migrations.RunPython` to take as an argument. Inside the function, you'd need to fetch a workable version of `Attribute` with `Attribute = apps.get_model("typeclasses", "Attribute")`, then get descs for Readables with like `descs = Attribute.objects.filter(db_key="desc", objectdb__db_typeclass_path="typeclasses.readable.readable.Readable")`.

Once you have those descs, you can start to build templates from them by looking at each Attribute's db_value, making a template from it if it's new, and then setting the Attribute's db_value to be `"[[TEMPLATE_%s]]" % new_template.id`.

If that sounds like too much to mess with, I've done that stuff a lot before so I wouldn't mind whipping it up, but if you wanna give it a whirl, by all means.

TehomCD: Probably has to be the Attribute db_strvalue field, though. Otherwise the template's markup characters would probably make it into a pickled list or whatever.

emonical: I think we need to be careful about this, especially if we are going to support 'owners'.

We need to reconcile around who owned the first instance, if they are banned players, etc.

Structurally, I had thought of a fourth 'ACCESS' level, which is like 'UNASSIGNED', and is attached to a root PlayerAccount that has been built for the purposes of serving as an owner for that, for auditing.

However, I think seeing the system in flight first is a wiser idea.

---

I was thinking of building an admin tool that would do an audit of a typeclasses (arbitrarily's) desc characteristics and return a menu of name and desc, along with details regarding the individuals that have made copies of these duplicates.

Then, being able to, by command go 'reconcile'. The advantage of this approach is being able to continually audit as necessary. It does place a certain additional onus on staff, but, given the fact that it assigns access to player accounts, I think there is some value in waiting.

Taking some time to test the system first doesn't preclude the ability to run the appropriate migrations.

TehomCD: I was looking at this and it'd be more tricky to do this as a migration than I thought, because the serialization/deserialization steps that Evennia does probably wouldn't work in the migration itself, forcing us to do that manually - the descs are stored in `db_value`, not `db_strvalue`, which seems like a bug. 

So probably building an admin tool might be good approach, yeah. If you're satisfied with how things look now, I think I can go ahead and merge this and we can see how things shake out before thinking of how to convert existing books to using templates.

emonical: I'm down. :)

TehomCD (approved):

PR #11, opened 2018-11-11
Fix a display bug and add the ability for template owners to see if t…

emonical:
…hey are in use and delete ones that are not in use.

There appears to be a display bug related to non-distinct queries. Also, this fixes several issues with the display and provides the ability to delete a template if it has been created but not applied.

### DIFF
--- a/world/templates/template_commands.py
+++ b/world/templates/template_commands.py
@@ -29,6 +29,9 @@ class TemplateForm(Paxform):
 
     When an object with a markup tag is looked at, the markup resolves
     to the description associated with template.
+    
+    Once a template has been created, it cannot be changed. A template
+    can be deleted, however, if it has not been applied to an object.
 
     The @study command, however, will NOT resolve the template markup.
     """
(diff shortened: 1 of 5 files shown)


world/templates/template_commands.py
```diff
@@ -86,17 +90,38 @@ def func(self):
             if not args:
                 super(CmdTemplateForm, self).func()
             else:
-                template = self.find_template(args)
-                if template:
-                    self.caller.msg(self.display(template))
+                try:
+                    template = Template.objects.filter(id=args)[:1].get()
```
TehomCD, line 94:
```suggestion
                    template = Template.objects.get(id=args)
```
is a little more readable to me, though should be the same thing

world/templates/template_manager.py
```diff
@@ -8,8 +8,8 @@
 
 class TemplateManager(models.Manager):
     def accessible_by(self, char):
-        return self.get_queryset().filter(Q(owner=char.roster.current_account) | Q(templategrantee__grantee=char.roster) | Q(access_level="OP"))
+        return self.get_queryset().filter(Q(owner=char.roster.current_account) | Q(templategrantee__grantee=char.roster) | Q(access_level="OP")).distinct()
```
TehomCD, line 11: Yeah, I always have trouble remembering when not having `.distinct()` will produce duplicates, but usually happens with complex queries in which you could have multiple matches or when having joins that could also produce multiple matches.

PR #12, opened 2018-11-13
Fix issue with template escaping.

emonical:
This adds additional escaping for the special escapechars. May need more if additional conflicts with regex tags.


TehomCD:
Nice, thanks! Will merge with rebase.

PR #73, opened 2018-12-07
Buy Orders

Zironic:
I'm reasonably sure this works. It could probably do with some logging code just in case.


world/petitions/petitions_commands.py
```diff
@@ -434,7 +527,8 @@ def cancel_sale(self):
         if sale.owner != self.caller.player_ob.Dominion:
             raise self.BrokerError("You can only cancel your own sales.")
         sale.cancel()
-        self.msg("You have cancelled the sale.")
+        broker_type = sale.broker_type
+        self.msg("You have cancelled the %s." % (dict(BrokeredSale.BROKER_TYPES)[broker_type].lower()))
```
TehomCD, line 531: Just as an fyi, django auto-creates a method for getting the string display from a field's value if it has `choices` set: it'd be `sale.get_broker_type_display()` here. Annoyingly they don't provide a method for looking up the value from the equivalent string, though.

TehomCD:
Looks good to me! I might add a few lines to the test about selling to existing purchase orders, and maybe logging like you suggest, but I think I'll go ahead and merge this. Thanks a ton!

world/petitions/petitions_commands.py
Zironic: That would look a bit neater. I havn't touched Django and Python in a year or two, so I actually took a while to figure out how to print it at all.

PR #77, opened 2018-12-08
Some humanization for Inventory, Org and Model

Zironic:
I did not comma separate the leaderboard because pretty table doesn't support it by default and it would require me messing with the values before they're passed to pretty table. I did however set it to right align which should help a little bit.

Bank, Shop and Broker to go.


TehomCD:
Pretty nice! Might as well do their currency on hand in inventory too while you're at it.

Zironic:
I completely missed currency on hand. I'll add that in the next batch which will include the bank, donate, praise and work.

Zironic:
The bank code has no tests and I didn't feel up to writing them so you'll want to double check them so they won't break horribly. I also found out why I missed currency at hand, it's because it was inside a mixin, I'm not sure that's the entirely best place for a string formater to be.

Also one issue I ran into while doing this is that if you're running all tests at once, they break because the dompc gets things added/removed by different tests in different modules.

Zironic:
I went ahead and wrote a command to check +score <person> and +propriety <modifier>

commands/base_commands/roster.py
```diff
@@ -936,7 +937,51 @@ def display_recognition(caller, charob):
...
 
+    If a propriety mod would be appropriate to a character, it can be +requested with a justification, if it is reflective of current or past RP and is not in dispute whether it happened, such as a white journal or public statement that represents a societal stance that's listed. Characters wishing to overturn propriety mods can pursue @actions such as social campaigns to try to reverse public opinion.
+    """
+    key = "@propriety"
+    aliases = ["+propriety", "propriety","reputation","honorific","honorifics"]
```
TehomCD: Oh, don't add the aliases with/without `+`/`@` etc. Evennia does that by default now and if you add them both with and without, it causes the multimatch of presenting the command multiple times.

honorifics are a distinct model from propriety, so having `@propriety <honorific>` is misleading when you're matching proprieties by name. 

You could do both based on the cmdstring used (listing honorifics if the `honorific` alias is used) if you wanted to, but since honorifics are unique per person it might be better to just list them as name/person pairs.

Zironic: I'm finding the terminology a bit confusing honestly. The honorifics are a lot less useful to search for since they tend to be unique but I guess I could make a command to just print all honorifics and their owners.

rozzingit: `@propriety <propriety mod>` would make more sense terminology-wise. Having 'reputation' in there also might be misleading when `score/reputation` is its own thing. I don't think there's particular desire from people to see a giant list of honorifics, just cause they're all unique, although I could see browsing through once just for sheer curiosity's sake.

TehomCD: Yeah, it might be simplest to wipe the aliases, keep the current behavior, and use `<propriety name>` instead of `<honorific name>` in the help string if you like.

Zironic:
I made it plural just so it won't interfere with the propriety helpfile

rozzingit:
Oh hey I just noticed today that big numbers appear in the `messenger` code when people are mailing around money!

Zironic:
Some of that will get caught by the banking code. I havn't looked to see if the give/messenger commands lives somewhere else from banktransfers.

Zironic:
Hmmmm. Adding an alias to messenger turned out to be very complicated and probably best not attempted, that code is a bit of a maze and it's best not made even more labyrinthine.

commands/base_commands/social.py
```diff
@@ -984,7 +984,7 @@ def func(self):
             self.set_or_remove_retainer_ability(attr_name, attr_desc)
             return
         # get the first new messenger we have waiting
-        if "receive" in self.switches:
+        if "receive" or "recieve"in self.switches:
```
TehomCD:
```suggestion
        if self.check_switches(["receive", "recieve"]):
```
Need to have either be `in self.switches` or it's just checking for whether a string literal is Truthy. Because of how verbose/annoying that gets, I made the `check_switches` method as just a quick set comparison between `self.switches` and any iterable you pass to it.

You're right that messenger is a mess so I don't blame you for not wanting to mess with it, though.

commands/base_commands/roster.py
```diff
@@ -936,7 +937,51 @@ def display_recognition(caller, charob):
...
 
+    If a propriety mod would be appropriate to a character, it can be +requested with a justification, if it is reflective of current or past RP and is not in dispute whether it happened, such as a white journal or public statement that represents a societal stance that's listed. Characters wishing to overturn propriety mods can pursue @actions such as social campaigns to try to reverse public opinion.
+    """
+    key = "@Proprieties"
+    aliases = ["proprietors"]
```
TehomCD: Probably dump the alias entirely so we don't accidentally have it collide with ownership of a business somehow later on. Command looks good! Needs to be added to a cmdset though. Probably add it around [here](https://github.com/Arx-Game/arxcode/blob/stable_orphan/commands/cmdsets/standard.py#L189) would make sense, maybe?

commands/base_commands/social.py
```diff
@@ -2191,7 +2192,28 @@ def func(self):
         """Execute command."""
         caller = self.caller
         if not self.switches or self.check_switches(self.prestige_switches):
-            return self.get_queryset_for_prestige_table()
+            if not self.lhs:
+                return self.get_queryset_for_prestige_table()
+            else:
+                targ = caller.search(self.lhs)
```
TehomCD: This is nice! If you wanted to, you could make it generic to org or player by searching for the AssetOwner directly, and searching either via the org name or `player__player__username`. Longname would only be applicable in the character case though.

Zironic: I think organisations should probably get longnames too. Right now the Knights of Solace are just 'Solace' while the Faith are 'Faith of the Pantheon', it's very inconsistent.

TehomCD: We have long names for player characters because there's many situations where you want to only have their name be briefly mentioned rather than their full name. Does the same really apply to organizations? I guess in tables, maybe. It'd probably make more sense to have add a short name as a SlugField for better urls for org pages and use it in tables, I guess, and then use the current name as the verbose/full name.

Zironic: I know I keep trying to go 'org faith' and 'org explorers' and the game saying mean things to me about not being a member of such an org.

TehomCD: For command matching we just need to update some of those commands to use [this method](https://github.com/Arx-Game/arxcode/blob/stable_orphan/commands/mixins.py#L52) that @Tanookami wrote. It does a contains check first for strings, then tries to narrow it down on multimatches with iexact.

Zironic: Unless I'm missing something, I can't search for the AssetOwner directly because AssetOwners don't have names, I'd have to grab all the Assets, then grab all the owners, then filter the owners.

What I could do is first try to find a playerornpc with the name, then try to find an org with the name, then grab their assets.

TehomCD:
```
qs = AssetOwner.objects.filter(Q(player__player__username__icontains=self.lhs) | Q(organization_owner__name__icontains=self.lhs))
if len(qs) > 1:
    qs = qs.filter(Q(player__player__username__iexact=self.lhs) | Q(organization_owner__name__iexact=self.lhs))
if len(qs) > 1:
    self.msg("Please narrow results.")
    return
obj = qs.first()
if not obj:
    self.msg("Nothing found")
    return
```
Something like that, probably.

Zironic:
With the change of +score to +notable. Do we still want the ability to check the prestige of individual characters/orgs and if so, what should it say?

TehomCD:
I think no, for the time being - my understanding from the discussion last night is that they weren't super keen on displaying individual rankings. I know that there's places that already display that sort of thing (cron, seeing values under inventory), but those might be going away - I'm not really sure where we stand on those right now. So probably best to hold off on it and then it might be added later.

Zironic:
I think all this humanization should be in a good state?

TehomCD:
Thanks much! Merged.

PR #96, opened 2018-12-19
Matchmaker

Zironic:
Confirmations for the broker and a prototype matchmaker that could do with some prettification.


world/dominion/general_dominion_commands.py
```diff
@@ -2514,8 +2513,12 @@ def func(self):
                 if num >= max_p:
                     caller.msg("You already have the maximum number of proteges for your social rank.")
                     return
-                if not self.check_social_rank_difference(char):
+                rankcheck, diff = self.check_social_rank_difference(caller.db.char_ob,char)
```
TehomCD:
```suggestion
                rankcheck, diff = self.check_social_rank_difference(caller.char_ob,char)
```
Getting rid of the `.db.char_ob` references - characters made after this point won't even have the Attribute set, so have to use the property.

world/petitions/models.py
```diff
@@ -10,7 +10,87 @@
...
+    SPONSOR = 1
+    PROTEGE = 2
+    PLOT = 3
+    TIME = 4
+    MATCH_TYPES = ((MARRIAGE, "Marriage"),(SPONSOR,"Sponsor"),(PROTEGE,"Protege"),(PLOT,"Plot"),(TIME,"Time"))
+    playtime_start = models.PositiveIntegerField(default=0)
+    playtime_end = models.PositiveIntegerField(default=0)
+    prefer_weekend = models.BooleanField(default=False)
+    interests = models.ManyToManyField(Interest)
+    clues = models.ManyToManyField(Clue)
```
TehomCD: You'll wanna set `blank=True` for the M2M fields if they're not required, so there's not validation errors in django admin if we tried to edit something and left them blank. I normally use the string literal way of specifying models for fields (like `"character.Clue"`) because of trying to avoid having models files import one another, which can lead to circular import errors or random errors when trying to load models during django's startup process.

world/petitions/petitions_commands.py
```diff
@@ -8,9 +8,179 @@
...
+    -Viewing:
+        matchmaker
+        matchmaker/all
+    -Editing
+        matchmaker/active #Toggles on/off
+        matchmaker/blurb <text>
+        matchmaker/playtimes <start>,<end> #Assume EST
+        matchmaker/clues <clue1>,<clue2>
+        matchmaker/tags <tag1>,<tag2>
+        matchmaker/seeking <marriage>,<plot>,<protege>,<sponsor> #any
```
TehomCD, line 32: Some brief examples of use for the docstring would probably be a good idea here.

world/petitions/petitions_commands.py
```diff
@@ -8,9 +8,179 @@
...
+        my_ad = self.caller.dompc.wanted_ad
+        ads = WantedAd.objects.filter(active=True)
+        matches_list = []
+        matched_list = []
+        for ad in ads:
+            matches = ad.match(self.caller.dompc)
+            if matches and (ad.TIME in matches):
+                matches_list.append(matches)
+                matched_list.append(ad)
+        table = PrettyTable(["Name", "Gender", "Age", "Matches","Playtime", "Blurb"])
```
TehomCD: You may wanna use EvTable rather than PrettyTable due to line wrap issues. Otherwise you run into that pretty fast unless you do a lot of truncation of the columns. Line-wrap looks pretty hideous in PrettyTable and makes it hard to read, so it's not just a cosmetic issue, unfortunately.

I prefer using PrettyTable where I know there won't be wrap issues due to speed - EvTable starts to have marked performance issues when there's a good number of matches, like beyond a few dozen. That's why for a few commands I have EvTable when there's a relatively small number of rows, but otherwise use PrettyTable with heavy truncation for a large number of matches.

world/petitions/models.py
```diff
@@ -10,7 +10,87 @@
...
+    searchtags = models.ManyToManyField(SearchTag)
+    blurb = models.TextField()
+    marriage = models.BooleanField(default=False)
+    sponsor = models.BooleanField(default=False)
+    protege = models.BooleanField(default=False)
+    plot = models.BooleanField(default=False)
+    active = models.BooleanField(default=False)
+    owner = models.OneToOneField(PlayerOrNpc, related_name="wanted_ad")
+
+    def match(self, dompc):
```
TehomCD, line 44: I'm a little concerned about the execution time of match, since this is called on every active Wanted Ad and contains nested queries with matching search tags and clues. I think it'd be better to do matches as a complex query based on everything but social rank which can't be queried, then doing that manually at the end. Maybe it wouldn't be too bad once all the Wanted Ads are in memory as a SharedMemoryModel, but I wouldn't be surprised that this could cause the game to lock up for a minute or so the first time someone did 'mm' after a reload.

TehomCD (changes requested):
There's a few changes I like that I left as code comments: changing `.db.char_ob` to `.char_ob`, adding `blank=True` to the M2M relationships, and maybe optimizing the matchmaking process a bit by using a complex query rather than nested queries while iterating through them.

The other thing would be to make the wanted_ad delete if the character is put back on the roster or moved to Gone, maybe filtering them out if they're just in the Inactive roster.

world/petitions/petitions_commands.py
Zironic: I had been wondering why you were using two different tables.

world/petitions/models.py
Zironic: The execution time would depend mostly on how Django implements the many2many relationship between the ads and the tags/clues. If it does it in a problematic way it would be possible to just store them as a list of strings in a .db property and do a list comparison instead.

Zironic:
I presume the sensible place to add the cleanup code would be in the admin commands for rostering? What it does now is simply that the commands for displaying the ads filters for active=True before passing them on to match.

world/petitions/models.py
TehomCD: I think the best way is to do it as a different query distinct from matches, and then append them if the ad is in that queryset. Like:

```
my_ad = self.caller.dompc.wanted_ad
ads = WantedAd.objects.filter(active=True)
if not my_ad.plot:
    plot_matches = []
else:
    plot_matches = ads.filter(Q(plot=True) & Q(clues__in=my_ad.clues.all()) | Q(searchtags__in=my_ad.searchtags.all())).distinct()
```
In `models.py` you could pass plot_matches as an arg and change the plot check to:
```
if self in plot_matches:
    matches.append(self.PLOT)
```

You could do stuff with conditional queries and annotations, but I think that's the part that matters since a nested query for each row in the table would be pretty bad news, so extracting it to only be done once would probably be all that's needed.

Zironic: That should work.

TehomCD:
Yeah, for deleting their stale ad so it doesn't carry on to the next player of the character, you'd want to use [post_roster_cleanup](https://github.com/Arx-Game/arxcode/blob/stable_orphan/server/utils/arx_utils.py#L321).

TehomCD:
I won't have time to ever fix this up myself, but feel free to reopen a new one if you ever return and want to update it.

PR #148, opened 2019-05-28
Add npc_create to admin_domain cmd

Tanookami:
Allows an NPC (playerless) domain to be generated (by staff) beneath an existing organization.

Easier for staff to create entities for a dominion-level story.

Requires an existing org name or id, a name for the new npc group, and their social rank #. Also added unit tests, but only for this switch of the command.


TehomCD (approved):
Looks good to me! Thank you <3

PR #159, opened 2019-06-21
Add lore topics to world page

### LINKED ISSUE #156: Lore On Website
I think this is already on someone's list, but...! I noticed in surprise that there isn't actually a feature request in for it. So. The problem is: all of the IC lore information, both basic and advanced, is now no longer accessible on the website. This tends to be a hurdle for pretty much everyone I've talked to, most importantly people new or returning to the game. It makes it harder to direct people to starter info, whereas that all used to be more accessible on the game's website. Even more confusingly, the helpfile header on the website's World page actually is named "Lore Topics," but it's all OOC helpfile info now.

We really need to have a page on the website for the Lore files! I think it'd probably be best to have maybe all the IC info on that page that's currently on the World page: the map, the orgs, links to the recipes, etc. And then split off the more OOC helpfiles and command files and the like. Or just have them on the same page but in different sections? It feels like a really integral part of the game is basically getting hidden away in a command that's not particularly obvious and is, at best, minorly disorganized. It's definitely not newbie-friendly in any way, and there's no good way to point people to good starter files to get acclimated to theme.

**Additional context**
I'm also willing to help in any reorganization of information that might be helpful to happen! There are times I actually split up information from the same Theme Questions response across multiple sections of the wiki page because the topics touched on were so different. Sometimes having stuff from the same response split into different entries can be the best for consumability.

Tanookami:
Adds basic lists of lore categories and topics to website.

Knowledge is power.

Tehom is cute. Fixes #156

### DIFF
--- a/web/help_topics/templates/help_topics/list.html
+++ b/web/help_topics/templates/help_topics/list.html
@@ -44,7 +44,7 @@
 
 	<div class="row {% cycle 'bg-success' 'bg-info' as rowcolors %} text-center">
 	<button data-toggle="collapse" data-target="#Topics">
-	<h1 class="bg-primary">Lore Topics</h1></button>
+	<h1 class="bg-primary">Gameplay Topics</h1></button>
 	</div> <!--button-->
 	<div id="Topics" class="collapse in">
     {% for category in all_categories %}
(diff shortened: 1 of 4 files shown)


TehomCD:
You are so wonderful. Thank you so much!

Issue #164, opened 2019-06-23
Don't show clue list if you mistype clue number


rozzingit:
**Is your feature request related to a problem? Please describe.**
Right now when you use the `clue` command to look at a clue -- say `clue 1234`, if you input a number of a clue you don't actually have, it tells you: `No clue found by that ID.` Which is great! But then it also proceeds to list...your entire clue list. For some of us, that's a lot of sudden spam, and honestly you have to scroll up to even see what was wrong to begin with.

**Describe the solution you'd like**
I don't really see the utility in adding the clue list at all. Could the message be updated to something like: ``You don't have a clue with that ID. Check `clues` to see your clue list.``

codersquid:
This bugs me too. I made a change that has the error message be the same as the error when someone calls the clue/share command  on a missing id and removed the table display. See what you think.

TehomCD:
Resolved by #165

PR #165, opened 2019-06-23
Do not display clue table for missing clues #164

codersquid:

This removes the table display when the clue <id> command is called by a user who does not have the corresponding clue.

I agree with the request in #164. That's been bugging me too.


TehomCD:
Sounds good to me.

Issue #144, opened 2019-05-15
Theory Notifications


ghost:
**Is your feature request related to a problem? Please describe.**
Theories shared with other players spawn an unhelpful notification inform that doesn't give much information about which theory is added. For characters with more than a handful of theories, identifying the new theory becomes very difficult. 

The current message is "Vellichor has shared a theory with you." 

**Describe the solution you'd like**
Change the inform messaging to see the theory name and number. 

E.g., Vellichor has shared theory 499, "Don't Eat Magic Books," with you.

**Describe alternatives you've considered**
* Note the theory by number in the emit. ("Vellichor shared theory 408 with you.") 

* Highlight the new theory with a different colour. 

* Highlight updated theories with a new colour. 

**Additional notes**
Ticket #107  mentioned this issue, but as part of a broader request.

codersquid:
What about: Vellichor has shared theory "Don't Eat Magic Books"' with you. Use **@theories %s** to view it.

PR #167, opened 2019-06-23
Add title and theory number to informs for #144

codersquid:

This adds the theory title and number to the inform when someone shares a theory.

#144 is a papercut for me too and the change to add it is pretty small. I figured, why not.

Maybe the description could also be included in the inform, but I didn't do it for this change.

This isn't ready to merge yet. There are no unit tests for theories that I could find, and I'd like to make at least one. I also haven't smoke tested this in a live environment. I'm having trouble populating my local db with characters.

### DIFF
--- a/web/character/investigation.py
+++ b/web/character/investigation.py
@@ -1658,7 +1658,8 @@ class CmdTheories(ArxPlayerCommand):
                     continue
                 theory.share_with(targ)
                 self.msg("Theory %s added to %s." % (self.lhs, targ))
-                targ.inform("%s has shared a theory with you." % self.caller, category="Theories")
+                targ.inform("%s has shared theory {w'%s'{n with you. Use {w@theories %s{n to view it." % (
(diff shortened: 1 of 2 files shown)


codersquid:
I've added a failing unit test. It follows what the other command tests are doing, but does not work. If I get around to it, I'll fix it. Meanwhile, if anything pops out as obviously wrong, leave a comment for me.

TehomCD:
At a guess, the blank message is probably because the feedback message is being sent to a different object (Character vs Account) than the caller is set to, and isn't being echo'd to them and caught by the mock for the `.msg` method. I forget how I normally resolve that - I might take the blank message and mock the other object's `.msg` method and check that, maybe.

codersquid:
I changed the caller and also mocked the Account inform vs the character one.

codersquid:
Hiya, I stood up a local instance to check this live and it seems all good.

TehomCD:
Looks good to me. Thanks!

PR #177, opened 2019-07-22
Matchmaker

Zironic:
This should contain a very basic matchmaker and whatever humanisation wasn't in the last pull.


world/petitions/petitions_commands.py
```diff
@@ -7,12 +7,188 @@
...
+        table.reformat_column(0,width=20)
+        self.msg(table)
+
+            
+    def clues(self):
+        ad = self.caller.dompc.wanted_ad
+        clues = []
+        args = self.args.split("/")
+        message="You're now matching: "
+        if args<1:
```
TehomCD:
```suggestion
        if not args:
```
A `len(args) < 1` would also work, but I think checking for falsy is usually more elegant

world/petitions/petitions_commands.py
```diff
@@ -7,12 +7,188 @@
...
+            ad.clues.clear()
+            self.msg("You're now matching no clues")
+            return
```
TehomCD: Maybe instead set message to "You're now matching no clues" and drop the return statement

world/petitions/petitions_commands.py
```diff
@@ -7,12 +7,188 @@
...
+        else:
+            ad.clues.clear()
+            for entry in args:
+                try:
+                    clue = self.get_by_name_or_id(Clue,entry)
+                except:
```
TehomCD: Please add an exception type - bare exceptions are kinda dangerous. Clue.DoesNotExist and TypeError/ValueError should catch anything you need

world/petitions/petitions_commands.py
```diff
@@ -7,12 +7,188 @@
...
+                    self.msg("There's no clue called {}".format(entry))
+                    self.msg(message)
+                    return
```
TehomCD: Might flow a bit better to set message here, drop the return, and make what would have followed into it an else block

world/petitions/petitions_commands.py
```diff
@@ -7,12 +7,188 @@
...
+                    ad.searchtags.add(tag)
+                message+="{}; ".format(clue.name)
+        self.msg(message)
+        return
+    
+    def tags(self):
+        ad = self.caller.dompc.wanted_ad
+        tags = []
+        message="You're now matching: "
+        if self.lhslist<1:
```
TehomCD:
```suggestion
        if not self.lhslist:
```
Reads a little clearer

world/petitions/petitions_commands.py
```diff
@@ -7,12 +7,188 @@
...
+            ad.searchtags.clear()
+            self.msg("You're now matching no tags")
+            return
+        else:
+            ad.searchtags.clear()
+            for entry in self.lhslist:
+                try:
+                    tag = self.get_by_name_or_id(SearchTag,entry)
+                except:
```
TehomCD: As above with the bare exception needing types and then preferring an else block to return

world/petitions/petitions_commands.py
```diff
@@ -7,12 +7,188 @@
...
+            return self.display_matches()
+        if "all" in self.switches:
+            return self.display_all_matches()
+        if "playtimes" in self.switches:
+            if len(self.lhslist) != 2:
+                self.msg("You need to supply both the start and end")
+                return
+            else:
+                ad.playtime_start=int(self.lhslist[0])
+                ad.playtime_end=int(self.lhslist[1])
```
TehomCD: You'll wanna catch ValueError here for non-numbers passed in

world/petitions/petitions_commands.py
```diff
@@ -7,12 +7,188 @@
...
+                        message+="marriage, "
+                    if entry.lower()=="plot":
+                        ad.plot=True
+                        message+="plot, "
+                    if entry.lower()=="sponsor":
+                        ad.sponsor=True
+                        message+="sponsor, "
+                    if entry.lower()=="protege":
+                        ad.protege=True
+                        message+="protege, "
```
TehomCD: I'd appreciate if all the PEP8 violations here were cleaned up, since it makes things a little hard to read and will make warnings harder to notice when inspections are turned on. Maybe run this through flake8?

TehomCD (changes requested):

world/petitions/petitions_commands.py
Zironic: I'll run it through autopep8

Zironic:
I believe I addressed all requests.

commands/base_commands/social.py
```diff
@@ -2213,12 +2213,14 @@ class CmdSocialNotable(ArxCommand):
...
     locks = "cmd:all()"
 
+
     def show_rankings(self, title, asset_owners, adjust_type, show_percent=False):
         counter = 1
         table = EvTable()
         table.add_column(width=8)
         table.add_column()
 
+
```
TehomCD: Please only have 1 line between methods/class variables, not 2.

commands/base_commands/tests.py
```diff
@@ -469,6 +469,10 @@ def test_cmd_give(self):
         self.assertEqual(self.assetowner2.economic, 50)
         self.account2.inform.assert_called_with("Char has given 50 economic resources to you.", category="Resources")
 
+
+    
+
+
```
TehomCD: snip these, please

world/dominion/general_dominion_commands.py
```diff
@@ -2537,8 +2536,12 @@ def func(self):
                 if num >= max_p:
                     caller.msg("You already have the maximum number of proteges for your social rank.")
                     return
-                if not self.check_social_rank_difference(char):
+                rankcheck, diff = self.check_social_rank_difference(caller.char_ob,char)
+                if not rankcheck: 
+                    self.msg("Your social rank must be at least %d higher than your target." % diff)                                      
                     return
+                else:
+                    pass
```
TehomCD, line 2544: Was this a placeholder you forgot to remove?

world/petitions/migrations/0004_auto_20181219_1343.py
```diff
@@ -0,0 +1,55 @@
...
+            options={
+                'abstract': False,
+            },
+        ),
+        migrations.AddField(
+            model_name='interest',
+            name='wantedads',
+            field=models.ManyToManyField(related_name='wanted_ads', to='petitions.WantedAd'),
+        ),
+    ]
```
TehomCD, line 55: This or your other PR will need to be changed when one or the other is merged in, because the petitions will have shared dependencies.

world/petitions/models.py
```diff
@@ -10,6 +10,93 @@
...
+    blurb = models.TextField()
+    marriage = models.BooleanField(default=False)
+    sponsor = models.BooleanField(default=False)
+    protege = models.BooleanField(default=False)
+    plot = models.BooleanField(default=False)
+    active = models.BooleanField(default=False)
+    owner = models.OneToOneField(
+        PlayerOrNpc,
+        related_name="wanted_ad",
+        on_delete=models.CASCADE)
```
TehomCD, line 46: This looks good. Intuitively I would have thought of it as a many-to-one with the booleans being a choices field, but this seems sensible

world/petitions/models.py
```diff
@@ -10,6 +10,93 @@
...
+            if rankcheck:
+                matches.append(self.SPONSOR)
+        if self.marriage and seeker_ad.marriage:
+            matches.append(self.MARRIAGE)
+        if self.plot and seeker_ad.plot:
+            intersect_clues = self.clues.all().intersection(seeker_ad.clues.all())
+            intersect_tags = self.searchtags.all().intersection(seeker_ad.searchtags.all())
+            if intersect_clues or intersect_tags:
+                matches.append(self.PLOT)
+        return matches
```
TehomCD, line 108: This could be very, very expensive if it was called in a loop. Ideally the match() should be replaced by a single query for matching clues and search tags if the one doing the query has plot up. I do django stuff in my day job so I don't mind handling this if it's not something you want to mess with.

world/petitions/models.py
```diff
@@ -21,19 +108,38 @@ class BrokeredSale(SharedMemoryModel):
...
+        "dominion.PlayerOrNpc",
+        related_name="brokered_purchases",
+        through="PurchasedAmount")
+    crafting_material_type = models.ForeignKey(
+        "dominion.CraftingMaterialType",
+        null=True,
+        blank=True,
+        on_delete=models.CASCADE)
+    broker_type = models.PositiveSmallIntegerField(
+        default=SALE, choices=BROKER_TYPES)
```
TehomCD: As I mentioned in the other code review, I apologize for not mentioning it, but the one PEP8 bit I don't adhere to is the 80 character per line limit, preferring 120. I much prefer horizontal placement of arguments for functions/classes so it's easier to tell what's going on when scrolling through code.

world/petitions/petitions_commands.py
```diff
@@ -7,9 +7,204 @@
...
+        ads = WantedAd.objects.filter(active=True)
+        table = evtable.EvTable(
+            "Name/Gender/Age",
+            "Matches",
+            "Times",
+            "Blurb",
+            border="cells",
+            width=78)
+        for ad in ads:
+            match = ad.match(self.caller.dompc)
```
TehomCD: Yeah, the O(N) queries in ad.match could actually be pretty troublesome - even something like 50 wanted adds doing queries on search tags and clues might be noticeable. What I'd probably do is turn it into a manager method of WantedAd, like `WantedAd.objects.find_matches(self.caller.dompc)`. But again, if that's not something you want to mess with, don't worry about it, it wouldn't take me long to figure something out.

world/petitions/petitions_commands.py
```diff
@@ -7,9 +7,204 @@
...
+            ad.clues.clear()
+            message = ("You're now matching no clues")
+        else:
+            ad.clues.clear()
+            for entry in args:
+                try:
+                    clue = self.get_by_name_or_id(Clue, entry)
+                    ad.clues.add(clue)
+                    for tag in clue.search_tags.all():
+                        ad.searchtags.add(tag)
```
TehomCD: Another situation of db operations in loops, but this probably wouldn't be as problematic. Still, would be better to get all the tags collectively with like `SearchTag.objects.filter(id__in=[ob.id for ob in clues])` if you got all the clues beforehand, for example, then add them to the ad.

TehomCD (changes requested):

world/dominion/general_dominion_commands.py
Zironic: I can honestly not remember why I wrote it like that.

world/petitions/migrations/0004_auto_20181219_1343.py
Zironic: I only created the migration because Travis can't run without it. I think it probably makes the most sense for you to handle the migrations on your end.

world/petitions/models.py
Zironic: I thought about using choices but when you're dealing with non-exclusive options like this I think a bunch of booleans are the easiest way to handle them.

world/petitions/models.py
Zironic: Didn't we already swap to using a single query for matching the clues and search tags? Match() currently only does everything else.

world/petitions/petitions_commands.py
Zironic: Does it make a difference? Wouldn't the filter have to do all the same operations that this loop is doing? Either way I'll change it.

Zironic:
This branch is enough revisions out of date that I should just remake it.

Issue #80, opened 2018-12-09
Making Petitions great


Zironic:
Reading the MSB thread there seems to be a lot of people that want to get use of petitions but arn't.

The biggest issue appears to be one of visibility. There's no notification when new petitions are added or old petitions are updated and old petitions often lie around stale in a way that makes people unsure if they're still meant to be relevant or not. I think this could be improved in a few ways.

1.1) All players should be notified of unread/new petitions the same way they do with subscribed bbposts
1.2) This should be something you can toggle off if you don't want the notifications
1.3) Petitions could get categories so you can choose which categories to get notifications about
2) Petitions should be coloured/marked based on if they have been read or not.
3.1) Petitions should be coloured/marked based on how long time it was since they were updated.
3.2) Petition owners should be able to make a stale petition look new again by updating it
4) Petitions should be sorted by last update rather then ID

TehomCD:
These all sound good to me. Probably add a date created field to `PetitionPost` and then a last modified/updated field to `Petition`. The categories idea is great - not sure if those should use existing `Tag` or `SearchTag` or be something distinct. `PetitionParticipation` could get a boolean field for wanting notifications, and let people toggle them off on a per-petition basis, or set it initially based on a setting they have if they globally want to suppress them.

rozzingit:
I might suggest petitions auto-closing (or maybe, at least, no longer displaying on the petition list) after a certain amount of time with no activity. Maybe an "expired" status (rather than "closed") that the petition owner could hit petition/refresh on to bring it back to life. Because I think there are a lot of petitions that people just post and forget. Or, like, the PC literally gets rostered, goes inactive, etc. (Actually, would be good to auto-close petitions if someone is marked as Gone and maybe auto-expire if they go inactive or back on the roster.) Maybe expired petitions show up on the list for the petition owner but in red (or some other marker for people who don't use colors) just to indicate "this still exists but it's not visible to others." Maybe an automatic @inform just to be like "your petition has expired due to lack of activity, petition/refresh ## to put it back on the petition list, or petition/close ## to close it permanently."

General categories or tags I think would be good: training, patronage (for seeking both patrons and proteges), action (for looking for people who want to join your action), research (investigation help)...there are probably more. That's what I can think of off the top of my head!

wahoo2:
Wait, I'm confused, doesn't this just make petitions exactly like an @bb?

TehomCD:
[...]

PR #178, opened 2019-07-23
Petitions

Zironic:
New functionality based on https://github.com/Arx-Game/arxcode/issues/80


Zironic:
Things not done

1) I've not created an option to ignore a specific petition
2) I've not done anything related to categories or tags

Features included are:

- Notification and messages sent when new notifications are posted or have subposts
- Petitions are now sorted by date updated
- Petitions are now color coded by status
- Petitions are now properly marked read or unread
- Petitions are cleaned up when a character is rostered

Things to keep in mind:
1) Characters won't have a settings object until after they use the petition command at least once and until then they will not receive notifications.
2) There may or may not be weirdness with the dates on legacy petitions.
3) The cleanup code has no tests.  Horrible things might happen.
4) It's not easy to test notifications or colors. Things might look and behave funky.

commands/base_commands/tests.py
```diff
@@ -468,6 +468,23 @@ def test_cmd_give(self):
...
+                      "Maximum AP: 300  Weekly AP Gain: 150\n"
+                      "You are carrying (Volume: 0/100):\n"
+                      "Money: coins worth a total of 125,446.00 silver pieces\n"
+                      "Bank Account:           0 silver coins\n"
+                      "Prestige:               0  Resources         Social Clout: 0\n"
+                      "|__ Legend:             0  Economic: 5,446\n"
+                      "|__ Fame:               0  Military:     0\n"
+                      "|__ Grandeur:           0  Social:       0\n"
+                      "|__ Propriety:          0\nMaterials:")
+
```
TehomCD, line 472: Looks like this is an accidental copy of the same test below

server/utils/arx_utils.py
```diff
@@ -768,15 +770,13 @@ def __get__(self, instance, owner):
...
     """
     Gets the full url when given a partial, used for formatting links. For this to work
     properly, you should define your Site's url under the 'Sites' app in django's admin
     site.
     Args:
         url: A partial url from a few, like '/namespace/view/'
-
     Returns:
         A full url, like "http://www.example.com/namespace/view/"
     """
```
TehomCD, line 784: Most of the whitespace changes violate PEP8. Spacing the docstrings I don't mind, but module-scoped functions are 2 spaces, methods are 1 space

world/petitions/models.py
```diff
@@ -167,15 +173,41 @@ def display(self):
         """Gets string display of the amount purchased and by whom"""
         return "{} bought {}".format(self.buyer, self.amount)
 
+class PetitionSettings(SharedMemoryModel):
+    owner = models.ForeignKey("dominion.PlayerOrNpc", related_name="petition_settings")
```
TehomCD, line 173: Probably a OneToOne with how you're using it, or move the fields into PlayerOrNpc as a whole. We'll probably make a more broad character-based settings object that'd absorb this at some point

world/petitions/models.py
```diff
@@ -167,15 +173,41 @@ def display(self):
...
+        try:
+            participations=self.owner.petitionparticipation_set.all()
+            for petition_participation in participations:
+                petition_participation.subscribed=False
+                petition_participation.unread_posts=True
+                petition_participation.signed_up=False
+                if (petition_participation.is_owner):
+                    petition_participation.petition.closed=True
+                petition_participation.save()
+        except:
```
TehomCD: I don't really see a need to have this in an exception block at all. It doesn't look like any exceptions are expected to be raised.

world/petitions/models.py
```diff
@@ -264,6 +301,15 @@ def mark_posts_read(self, dompc):
...
         except PetitionParticipation.DoesNotExist:
             pass
+    def mark_posts_unread(self, dompc):
+        """If dompc is a participant, mark their posts read"""
+        try:
+            participants = self.petitionparticipation_set.all().exclude(dompc=dompc)
+            for participant in participants:
+                    participant.unread_posts = True
+                    participant.save()
+        except PetitionParticipation.DoesNotExist:
```
TehomCD: DoesNotExist is never raised by a filter/exclude statement, so no need for the exception block here

TehomCD (changes requested):
I'll review in more detail later

server/utils/arx_utils.py
Zironic: The whitespace changes are all artifacts of me cleaning up the merge and are only still there because they're super annoying to hunt down and delete.

world/petitions/models.py
Zironic: Yeah, this object exists just to make the settings modular and easy to clean up.

TehomCD (changes requested):
I apologize for being difficult, but although I appreciate you addressing the PEP8 changes for spacing, I should have said I go with line lengths of 120 rather than 80. I find the short line lengths with each argument on their own line makes it harder to scan a file, and I'd really prefer most of those to be changed back both for my own ease of reading the code and for a consistent format. Other than that it looks good to me.

Zironic:
It took some messing around with GIT but it should be 120 line length formatted now.

TehomCD (approved):
Looks good to me. Thanks!

PR #185, opened 2019-08-03
Fixed some bugs and improved color coder

Zironic:
Found and addressed some bugs in the color coder and made a stab at improving the maintainability of the function.

### DIFF
--- a/world/petitions/petitions_commands.py
+++ b/world/petitions/petitions_commands.py
@@ -9,7 +9,7 @@ from server.utils.exceptions import PayError, CommandError
 from server.utils.prettytable import PrettyTable
 from world.petitions.forms import PetitionForm
 from world.petitions.exceptions import PetitionError
-from world.petitions.models import BrokeredSale, Petition, PetitionSettings
+from world.petitions.models import BrokeredSale, Petition, PetitionSettings, PetitionParticipation
 from world.dominion.models import Organization
 from datetime import date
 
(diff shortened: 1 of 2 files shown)


TehomCD:
Maybe later we can add a unit test that calls the color method with petitions set to different values and asserts the strings are returned, but this looks fine for now.

rozzingit:
Looks like people are getting informs/notifications for petitions in orgs they're not in.

```
A new petition was posted by Mirk to Halfshav.
You have new informs. Use @inform 54 to read them.
**********************************************************************

Category: Petition
----------------------------------------------------------------------

A new petition was posted by Mirk to Halfshav.

ID: 720  Topic: Test
Owner: Mirk  Organization: Halfshav
Description: Test

Signups: 

**********************************************************************
```

Got that on both my bits, neither of which are in Halfshav.

Zironic:
Yeah, it turns out I completely forgot to add the access filter code. Tehom should have gotten it fixed now. Let me know if you discover any other issues or have any future requests regarding petitions.

rozzingit:
Looks like there might be issues with `petition/ignore`. Hitting `petition/ignore general` the first time returns: `You are now informed of new general petitions.` (To be clear, I definitely hadn't ignored them previously, because I had gotten a general petition inform shortly before that.) Which is probably the opposite of desired behavior, I assume! And my instinct is that it's probably meant to be a toggle, but I get the same response every time. (YOU SHALL BE INFORMED OF GENERAL PETITIONS WHETHER YOU LIKE IT OR NOT.)

PR #189, opened 2019-08-15
Add server config for cost of writing new clues

Tanookami:
Allows admin to set the action point cost of writing a new clue for an investigation. Also adds a filter in django admin to list the Clues that are missing search tags. Adds unit tests.

Replaces a hard-coded action point price. Lets admin easily see which clues are not currently able to be investigated due to lack of search tags.


TehomCD (approved):
Looks good to me. Thank you! <3

PR #192, opened 2019-08-23
Fix Flashback allow switch

Tanookami:
Removes the use of 'prefetch_related' when looking up flashback post reader permissions. Adds unit tests to check the changing views of a reader. Also fixes a UX text bug.
As prefetch_related caches results, instances of pre-existing reader permissions were not caught. This caused IntegrityError when attempting bulk creation of permissions that already exist.
addresses in-game ticket 17701


TehomCD (approved):
LGTM. Thank you! <3

PR #196, opened 2019-08-26
Cal plotroom/location toggle

### LINKED ISSUE #195: Clear Event Location or Plotroom During Creation
You can't submit an event if it has both a location and plotroom set. This makes sense! However, right now there's no way to _clear_ either of those fields if you accidentally set both. If a player accidentally sets both, they have to use `cal/abort` and restart the creation process. (And given that two different players have asked about it today, I'd say it's probably something that happens on occasion.) So basically: there needs to be some method to either automatically or manually be able to clear these fields during the event creation process.

I think the best way to do this would be for the `cal/location` command to clear an event's plotroom automatically when used, and for `cal/plotroom` to clear the event's location. An event should never have both set at once, so I don't foresee it causing an issue to have an automatic clearing. This way, a player might set a location, then go, oh oops I meant to use a plotroom instead, they set a plotroom, and the location is automatically cleared out. (And, of course, vice versa if they start with a plotroom but then realize an on-grid location is what they actually want.)

You could also make it so that `cal/location` and `cal/plotroom` by themselves (with no other input) would just clear the fields manually. However, I think the automated version would be better, and I'm not personally seeing potential problems that would be created from it. I could be wrong!

Tanookami:
When setting an event's location, its plotroom field will be set to 'None' and vice versa.
There's no way to nullify the location field in-game, forcing event creation from scratch.
Fast and easy solution, thanks! Closes #195

### DIFF
--- a/commands/base_commands/social.py
+++ b/commands/base_commands/social.py
@@ -1662,6 +1662,16 @@ class CmdCalendar(ArxPlayerCommand):
         if event:
             self.caller.char_ob.pay_money(cost)
 
+    def set_event_desc(self, event):
+        """Sets description of an event"""
+        self.set_form_or_event_attribute("desc", self.lhs, event)
+        self.msg("Desc of event set to:\n%s" % self.lhs)
+
+    def set_room_desc(self, event):
+        """Sets description appended to event's location"""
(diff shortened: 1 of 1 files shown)


TehomCD (approved):
LGTM. Thank you! <3

PR #198, opened 2019-09-07
Order petitions from old to new

### LINKED ISSUE #193: New Petitions On Bottom
Petitions used to be listed with new petitions on the bottom instead of the top. This matches the general pattern across the rest of the game, and I think is a good design philosophy in general for MU*s, because it avoids having to scroll up to get new information. (Which is, in turn, reverse from what's usually best for the web. But different mediums!)

Since the updates from #178 (which are awesome!!), new petitions show up at the top of the petition list. This is especially confusing because the petitions on the list from before the changeover are still in the same order. So it goes New-Old-Old-New.

Can we have petitions sorted old to new again?

Tagging @Zironic because I believe this was his PR!

codersquid:
added default ordering to the Petitions model to fix #193

### DIFF
--- a/world/petitions/models.py
+++ b/world/petitions/models.py
@@ -201,6 +201,9 @@ class Petition(SharedMemoryModel):
     date_created = models.DateField(auto_now_add=True)
     date_updated = models.DateField(auto_now=True)
 
+    class Meta:
+        ordering = ('-date_updated',)
+
     @property
     def owner(self):
         """Gets first owner, if any"""


codersquid:
I just now made this in a :sleeping: haze and realize maybe I did this in haste because I notice in a few places in the petitions_command the query is just all(), but on other places it's explicitly ordered by date updated. hold a sec while I get coffee and ask for the preference.

codersquid:
gonna squash and fix this to order by the date updated.

TehomCD (approved):
Looks good to me, thanks! I'll go ahead and generate the migration that django insists on creating for any Meta change.

PR #212, opened 2019-10-07
Mark investigation inactive when preemptively learning its clue

Tanookami:
When a clue discovery occurs, we'll now mark our investigations that targeted it as inactive instead of just removing their clue_target. This forces a new clue_target acquisition if the user chooses to mark it active again. Also adding a unit test for it.
Simply removing the target bypassed the "no clueless investigations" admin setting.


TehomCD (approved):
Excellent work, thank you. :heart:

PR #244, opened 2019-12-15
Donate Help Update v2

Sritterbush:
Second run of testing my contribution ability.

This should modify the help file so that any future changes to the AP cost to run the Donate command automatically is reflected in the help file. Not a feature anyone asked for, still testing the waters to ensure I'm doing this right.

### DIFF
--- a/commands/base_commands/social.py
+++ b/commands/base_commands/social.py
@@ -2618,12 +2618,19 @@ class CmdDonate(ArxCommand):
         +donate/score [<group>]
 
     Donates money to some group of npcs in exchange for prestige.
-    +donate/score lists donation amounts. Costs 1 AP.
     """
     key = "+donate"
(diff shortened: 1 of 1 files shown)


TehomCD:
Okay so this is giving a merge conflict because you're using an older version of the `stable_orphan` branch, before your previous changes were merged in. Since this modifies the same lines of code your previous PR did, it sees it as two separate commits trying to change the same lines of code, and can't resolve that automatically. To resolve it, it needs the previous commit you applied, so it knows that this change comes later. You want to update your fork, either via git (adding the main repo as a remote, pulling changes from it, pushing to your fork) or by going to your fork's github page and selecting an option to sync up for you (I think it'll be a button/popup somewhere on there). Then you make your new branch as normal, and it'll have the changes you had applied previously, preventing a merge conflict.

Issue #236, opened 2019-11-28
Shop Requiring Full Price Despite Discount


Cobaltasaurus:
+shop/buy is now requiring that you have the full cost of an item to let you buy it, but will still only charge you the discount price.

Example:

Josephine put a jewelry piece on her shop for a full price of 20,000, and Mirari had a discount of 15%, meaning she should pay 17,000.  With 17,050 silver in her inventory +shop/buy would not let her buy the item.  When she had 20,000 in her inventory +Shop allowed her to buy it at the 17,000 price.

Sritterbush:
This edit should fix this problem.

`
    def pay_owner(self, price, msg):
        """Pay money to the other and send an inform of the sale"""
        loc = self.caller.location
        loc.db.shopowner.pay_money(-price)
        assets = loc.db.shopowner.player_ob.assets
        if price - price * (self.get_discount() / 100.0) >= assets.min_silver_for_inform:
            assets.inform(msg, category="shop", append=True)
`

Sritterbush:
This check is probably caused by the if price >= assets.min_silver_for_inform check in def pay_owner. 

Proposed new code to correct this involves altering line 1124 of home.py

`        if price - price * (self.get_discount() / 100.0) >= assets.min_silver_for_inform:`

Sritterbush:
Will test this in game after I've confirmed the game has the latest copy of the github on it, and determine if the fix worked.

Sritterbush:
Tested, confirmed fixed.

Issue #231, opened 2019-10-30
Shop Blocking Org Public Affiliation Only


SparkleDot:
Currently if you blacklist an org from your shop, it appears that it blocks both public and secret affiliations to those blocked orgs. This doesn't seem fair to secret affiliations and they shouldn't be blocked unless it's public. My shopkeepers aren't the Inquisition, we don't know your private dealings.

The solution I'd like to see is that the shop code checks public membership of a blocked org, and if true, will stop that person from being able to buy. However, if false, then those people can still use that shop while secretly being part of that org.

Sritterbush:
Altering line 1154 of home.py in the following way should fix this.

`      if org.name in blacklist and caller in org.name = [ob.player.player for ob in org.active_members if not ob.secret]:`

PR #247, opened 2019-12-17
donate help v4

Sritterbush:
Added the caller reference, which was missing from the previous pull review.

### DIFF
--- a/commands/base_commands/social.py
+++ b/commands/base_commands/social.py
@@ -2625,7 +2625,6 @@ class CmdDonate(ArxCommand):
     action_point_cost = 1
 
     def get_help(self, caller, cmdset):
-        caller = caller.player_ob
         msg = self.__doc__ + """
     +donate/score lists donation amounts. Costs {w%s{n AP.
     """ % (action_point_cost)
--- a/commands/base_commands/social.py
+++ b/commands/base_commands/social.py
@@ -2618,19 +2618,12 @@ class CmdDonate(ArxCommand):
         +donate/score [<group>]
 
     Donates money to some group of npcs in exchange for prestige.
+    +donate/score lists donation amounts. Costs 1 AP.
     """
     key = "+donate"
     locks = "cmd:all()"
     help_category = "Social"
     action_point_cost = 1
-
-    def get_help(self, caller, cmdset):
-        msg = self.__doc__ + """
-
-   +donate/score lists donation amounts. Costs {w%s{n AP.
-
-    """ % (action_point_cost)
-        return msg
 
     @property
     def donations(self):
--- a/commands/base_commands/social.py
+++ b/commands/base_commands/social.py
@@ -2618,12 +2618,19 @@ class CmdDonate(ArxCommand):
         +donate/score [<group>]
 
     Donates money to some group of npcs in exchange for prestige.
-    +donate/score lists donation amounts. Costs 1 AP.
     """
     key = "+donate"
     locks = "cmd:all()"
     help_category = "Social"
     action_point_cost = 1
+
+    def get_help(self, caller, cmdset):
+        msg = self.__doc__ + """
+
+   +donate/score lists donation amounts. Costs {w%s{n AP.
+
+    """ % (action_point_cost)
+        return msg
 
     @property
     def donations(self):
--- a/commands/base_commands/social.py
+++ b/commands/base_commands/social.py
@@ -2625,6 +2625,7 @@ class CmdDonate(ArxCommand):
     action_point_cost = 1
 
     def get_help(self, caller, cmdset):
+        caller = caller.player_ob
         msg = self.__doc__ + """
     +donate/score lists donation amounts. Costs {w%s{n AP.
     """ % (action_point_cost)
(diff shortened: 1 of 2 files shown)


commands/base_commands/social.py
```diff
@@ -2618,13 +2618,19 @@ class CmdDonate(ArxCommand):
...
     Donates money to some group of npcs in exchange for prestige.
-    +donate/score lists donation amounts. Costs 1 AP.
     """
     key = "+donate"
     locks = "cmd:all()"
     help_category = "Social"
     action_point_cost = 1
 
+    def get_help(self, caller, cmdset):
+        caller = caller.player_ob
```
TehomCD: You don't need this line here, since `caller` isn't used for the rest of the function. Unused local variables fall into the category of 'linting errors' or warnings: things which won't cause any errors in code, but are things you should avoid. A lot of code editors will flag them to let you know to fix them.

commands/cmdsets/home.py
```diff
@@ -1121,7 +1121,7 @@ def pay_owner(self, price, msg):
         loc = self.caller.location
         loc.db.shopowner.pay_money(-price)
         assets = loc.db.shopowner.player_ob.assets
-        if price >= assets.min_silver_for_inform:
+        if price - (price * (self.get_discount() / 100.0)) >= assets.min_silver_for_inform:
```
TehomCD: I believe the price here is already discounted. Did you mean `+` rather than `-`, so it'd use the undiscounted price for sending an inform? I'm not sure I follow what the intent is here. 

A best practice is often to make different commits and even separate pull requests for unrelated changes: it makes it easier to see what your intent is during code reviews and helps in version control to identify when things changed and how.

TehomCD:
Oh one thing that is helpful, if you don't mind, is to link to issues that you're addressing in a PR in the description or the commits. If you write, say, `Closes #236 ` in a commit message, it'll automatically close issue 236 if the PR is merged. In github, writing the pound sign followed by an issue number will create a link in that issue back to where it's referenced.

Thank you for your work in making PRs! These are looking very good for someone so new to it.

commands/base_commands/social.py
Sritterbush: Ah, I see! I thought it would be alright without, but talking to someone else gave me the impression I should include it. My confusion entirely.

commands/cmdsets/home.py
Sritterbush: I didn't realize this pushed through. I was testing an edit based on a bug request, didn't think it would append to my existing pull request.

TehomCD: Ah! Okay, so the deal is that every pull request is tied to a branch: when you make a pull request, you're comparing the set of changes of commits on one branch to a branch you want to merge them into. Once the PR is opened, any further commits on that branch are added to it. To make a separate pull request, you'd make a different branch with the new set of changes.

Sritterbush: So while I have a pending pull request in, all the updates to my own git automatically alter my pull request in process. I didn't know that before, but good info. I'll be mindful of this moving forward.

Sritterbush:
Included an adjustment to the price code that should resolve  the issue in issue 236 where a shop asks you to have the silver on hand needed to make the purchase, even when you have a discount. This check appears to not factor in the discount because it happens before the buy command is validated.

Closes #236

Sritterbush:
Reviewed the edit after your comment, realized I hit the code in the wrong place. Moved the edit to what I now believe is the correct price reference, in line 1199. I am... Very new to this. Not used to trying to test code without running it.

commands/cmdsets/home.py
TehomCD: Well, on that branch: you can have multiple branches coexisting simultaneously and switch between them. It'd be like:
```
git checkout stable_orphan
git pull upstream
git checkout -b fix-something
git commit -am "Fixed thing!!"
git push -u origin fix-something
git checkout stable_orphan
git checkout -b fix-other-thing
git commit -am "Fixed other thing!"
git push -u origin fix-other-thing
```
That'd get you two independent branches with different histories/changes.

commands/cmdsets/home.py
```diff
@@ -1151,7 +1151,7 @@ def check_blacklist(self):
         if caller in blacklist:
             return True
         for org in caller.player_ob.Dominion.current_orgs:
-            if org.name in blacklist:
+            if org.name in blacklist and caller in org.name = [ob.player.player for ob in org.active_members if not ob.secret]:
```
Sritterbush: Okay, so I had a /third/ edit in process I didn't remember I was looking at. FML. This was in response to issue #231, but I wasn't sure it was a valid solution yet. I don't quite understand the code behind it.

TehomCD (commented):
You're doing just fine. :) I have to go to bed, so I'll review stuff tomorrow. Thanks again for your work!

commands/cmdsets/home.py
Sritterbush: Reverted this edit, it failed the Travis CI

TehomCD:
Looks good to me. Thanks! Remember to pull changes after this is merged to update your local repository.

commands/cmdsets/home.py
Sritterbush: I'm not seeing that flow of commands, since I'm using Github Desktop. I have forsaken my command prompt origins.

TehomCD:
The easiest thing to do would be go to your fork on GitHub, hit the button to sync it up, then pull the changes locally. Then branches from that point will be up to date.

PR #248, opened 2019-12-18
Fixing Help Donate (Because I broke it)

Sritterbush:
Adding the self.* reference to the line that is breaking code, because I need to remember my internal variables like the flag that they're internal in python.


TehomCD:
Looks good to me. Thanks!

PR #250, opened 2019-12-28
Change flashback navigation

codersquid:
Change the 'Back to' navigation to go to flashbacks instead of the character page.

This changes the flashback page template to link back to the flashback list rather than the character page.

With multiple ongoing flashbacks I find that I would like to have a link back to the list of flashbacks rather than the character page. That is less clicking and scrolling.


TehomCD:
Thank you for this!

Issue #232, opened 2019-11-01
Tag search in @clues


AVeryImpressivePerson:
A tag-only clue search would be very handy for weeding out false positives.

Sritterbush:
This would resolve issue #232

PR #252, opened 2020-01-02
New Feature: clues/tagsearch 

Sritterbush:
Adds a command to clue that allows a tagsearch function, distinct from search in that it only searches tags.

Edit adds the new functionality, updates the clue helpfile.


Sritterbush:
This would close issue #232

web/character/investigation.py
```diff
@@ -1461,11 +1463,14 @@ def share_clues(self):
...
         discoveries = self.clue_discoveries.select_related('clue').order_by('date')
+        if "tagsearch" in self.switches:
+            msg = "{wMatching Clues{n\n"
+            discoveries = discoveries.filter(Q(clue__search_tags__name__iexact=self.args)).distinct()   
         if "search" in self.switches:
             msg = "{wMatching Clues{n\n"
             discoveries = discoveries.filter(Q(message__icontains=self.args) | Q(clue__desc__icontains=self.args) |
                                              Q(clue__name__icontains=self.args) |
-                                             Q(clue__search_tags__name__iexact=self.args)).distinct()
+                                             Q(clue__search_tags__name__iexact=self.args)).distinct()
```
TehomCD: Finally getting to this after I finished the python 2->3 change. Sorry for making you wait for so long! Looks good to me, though I'd like you to remove the trailing whitespace at the end of line 1473. While it's not a huge deal I'd prefer to not put in any linting errors until they're handled automatically by Black or something
```suggestion
                                             Q(clue__search_tags__name__iexact=self.args)).distinct()
```

Selerik:
Been swamped, will get the code cleaned tonight and resubmitted. May have a
more elegant way to include it that allows searching by other specific
fields, rather than only by tag, as well.

web/character/investigation.py
```diff
@@ -1381,7 +1384,7 @@ def clue_discoveries(self):
     def func(self):
         """Executes clues command"""
         try:
-            if not self.args or "search" in self.switches:
+            if not self.args or "search" in self.switches or "search/tag" in self.switches or "search/name" in self.switches:
```
TehomCD: I don't think this will work - I believe that switches split based on the "/" separator, so for those it would be that "search" is in self.switches and "tag" is in self.switches. I'll need to double check that

dvoraen: Belatedly, but that seems to be correct (needing to check for "B in self.switches and C in self.switches" after confirming A).  Example of this is in the flashback code with "flashback/check/flub" effectively doing that.  (It checks for ("check","roll") first, then calls a function which checks for "flub" on its own.  See scene_commands.py lines 69 and 255.)

Because of that, I don't think it's necessary to even look for /tag or /name in this particular if, since they're subordinate to /search.

```suggestion
            if not self.args or "search" in self.switches:
```

web/character/investigation.py
```diff
@@ -1461,6 +1464,12 @@ def share_clues(self):
     def disp_clue_table(self):
         table = PrettyTable(["{wClue #{n", "{wSubject{n", "{wType{n"])
         discoveries = self.clue_discoveries.select_related('clue').order_by('date')
+        if "search/tag" in self.switches:
```
TehomCD:
```suggestion
        if "tag" in self.switches:
```

web/character/investigation.py
```diff
@@ -1461,6 +1464,12 @@ def share_clues(self):
     def disp_clue_table(self):
         table = PrettyTable(["{wClue #{n", "{wSubject{n", "{wType{n"])
         discoveries = self.clue_discoveries.select_related('clue').order_by('date')
+        if "search/tag" in self.switches:
+            msg = "{wMatching Clues{n\n"
+            discoveries = discoveries.filter(Q(clue__search_tags__name__iexact=self.args)).distinct()
+        if "search/name" in self.switches:
```
TehomCD:
```suggestion
        elif "name" in self.switches:

```

web/character/investigation.py
```diff
@@ -1461,6 +1464,12 @@ def share_clues(self):
     def disp_clue_table(self):
         table = PrettyTable(["{wClue #{n", "{wSubject{n", "{wType{n"])
         discoveries = self.clue_discoveries.select_related('clue').order_by('date')
+        if "search/tag" in self.switches:
+            msg = "{wMatching Clues{n\n"
+            discoveries = discoveries.filter(Q(clue__search_tags__name__iexact=self.args)).distinct()
+        if "search/name" in self.switches:
+            msg = "{wMatching Clues{n\n"
+            discoveries = discoveries.filter(Q(clue__name__icontains=self.args)).distinct()
         if "search" in self.switches:
```
TehomCD:
```suggestion
        elif "search" in self.switches and self.args:

```

TehomCD (approved):
Got around to fixing this

PR #267, opened 2020-01-30
fix a display string

Tanookami:
Prevents a null setting from causing AttributeError due to not having a .capitalize() method.
Fix bug!


TehomCD (approved):
Very nice! <3

🚑

PR #275, opened 2020-03-07
fix db lock inv bug

ApostateCD:
Database locks would make wear fail by having a null value in worn time, fixed to have a 0 in case of db locks


TehomCD:
Grats on your first PR! 👽

Tanookami (approved):
Looks fine.

Issue #194, opened 2019-08-24
Petition/Ignore Bug


rozzingit:
Since the fantabulous updates to petitions from #178 by @Zironic, there's actually a bug in the `petition/ignore` switch. That is -- it doesn't work! If you try the `petition/ignore general` or `petition/ignore <org>` options, it tells you "You are now informed..." every time so that there's no way to turn off those inform options. (If you do `petition/ignore all` it does give you feedback that it's turning the informs off.)

I SUPER appreciate having informs now, so many thanks for that. Just seems there's probably a minor bug somewhere.

Zironic:
This bug isn't what it looks like. The ignore command itself works just fine however I completely botched the text output.

Here is a fix
 ```
   def do_ignore_switches(self):
        settings, created = self.caller.dompc.petition_settings.get_or_create()
        if self.lhs == "all":
            settings.inform ^= True
            self.msg("You are now %sinformed of new petitions." % ("" if settings.inform else "not "))
        elif self.lhs == "general":
            settings.ignore_general ^= True
            self.msg("You are now %sinformed of new general petitions." % ("" if settings.ignore_general else "not "))
        else:
            org = self.get_org_from_args(self.lhs)
            if org in settings.ignored_organizations:
                settings.ignored_organizations.remove(org)
                msg = "You are now informed of new "
            else:
                settings.ignored_organizations.add(org)
                msg = "You are now not informed of new "
            self.msg(msg + self.lhs + " petitions.")`
```

**Comment by rozzingit** [comment]
Apparently the option to ignore ALL petitions would give the right feedback, but it still wouldn't actually stop the petition informs. So idk if it's just the text output.

**Comment by Zironic** [comment]
I think I may have figured out why. 
```   
     def do_ignore_switches(self):
        settings, created = self.caller.dompc.petition_settings.get_or_create()
        if self.lhs == "all":
            settings.inform ^= True
            self.msg("You are now %sinformed of new petitions." % ("" if settings.inform else "not "))
            settings.save()
        elif self.lhs == "general":
            settings.ignore_general ^= True
            self.msg("You are now %sinformed of new general petitions." % ("" if not settings.ignore_general else "not "))
            settings.save()
        else:
            org = self.get_org_from_args(self.lhs)
            if org in settings.ignored_organizations:
                settings.ignored_organizations.remove(org)
                msg = "You are now informed of new "
            else:
                settings.ignored_organizations.add(org)
                msg = "You are now not informed of new "
[...]

PR #277, opened 2020-03-15
Petition Informs fix

### LINKED ISSUE #215: Org Petitions Going to Non-Org Members
It seems that people are sometimes getting informs for org petitions for orgs they're not members of. Some Whisper players were testing Whisper petitions tonight for one member who _wasn't_ getting informs, only to hear from someone (who, granted, was once a member of the org, but not for a good long while now) outside the org that they had seen the petition informs. So: very odd! My PC is a member of the org and also received the informs, but is not actually at an org rank that can _view_ the informs. So I'm not sure if that's an issue with the informs needing to be locked to certain ranks based on the org perm or if it's similar to the issue with people getting petition informs for orgs they're not in.

Additionally, some people who _are_ org members aren't getting petition informs for their org at all. Idk!

Tagging @emonical just cause I know the petition code is his.

Tanookami:
In the Petition submit command, excludes deguilded org members from being informed. Also adds a test. And finally, setting up Dominion features for new characters will quietly create PetitionSettings for them, so they receive informs by default.
Fixes a bug and adjusts petition inform behavior to act more intuitively
Resolves #215

### DIFF
--- a/paxforms/paxform_commands.py
+++ b/paxforms/paxform_commands.py
@@ -9,7 +9,8 @@ class PaxformCommand(ArxCommand):
     def __init__(self):
         super(ArxCommand, self).__init__()
         self._form = self.__class__.form_class()
-        if not self.__doc__ or len(self.__doc__) == 0:
+        doc = self.__doc__
+        if not doc or len(doc) == 0 or doc == ArxCommand.__doc__:
             self.__doc__ = self.__docstring
         self._extras = None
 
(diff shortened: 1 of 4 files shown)


TehomCD (dismissed):
Awesome! Thank you so much <3

TehomCD (approved):
🥂

Zironic:
I think that while you're doing this, you should probably fix the bug in the petition settings themselves:
https://github.com/Arx-Game/arxcode/issues/194

As you can see I already wrote the code, I'm just too lazy to write tests for it and make a PR.

PR #280, opened 2020-03-25
override twisted err message in scripts

Tanookami:
Overrides twisted error message method to give a full traceback; to the log, not to the user.
The single line error is not helpful to trace script bugs.


TehomCD (approved):
This is such an elegant solution. Thank you!

PR #282, opened 2020-04-02
Fix a bug and add tests

Tanookami:
Fix a bug where Work command chooses the first job matching any skill, rather than one matching the highest skill a character possesses.
Bugs are contrite for causing problems.


TehomCD (approved):
LGTM, thank you for fixing this. ❤️

PR #290, opened 2020-04-19
Hotfix for shop recipe availability

LeBulldoge:
### Brief overview of PR changes/additions
Left an error where the caller was the one being checked for recipes, not the crafter, which meant that you couldn't craft anything from the shop unless you... could craft it.

### Motivation for adding to Arx
Bugfixing is fun, but fixing bugs that you pushed is even more fun.

### Other info (issues closed, discussion etc)
Overlooked it when developing and the tests were completely wrong, didn't cover the standard use case to begin with. Apologize for the inconvenience! ;w;


TehomCD (approved):
Oops, good catch. My bad for missing that

PR #291, opened 2020-04-22
Feature: Channel mentions

LeBulldoge:
Added a private method to the `Channel` class, which looks through a message for mentions of users starting with `@` and highlights them if found.
Fairly often there might be more than one ongoing conversation in any given channel (especially Info) and it's quite hard to tell who is talking to who, even if they mention their name directly. Using an `@` symbol for this purpose is probably akin to muscle memory for a lot of people.
The formatting is obviously subject to change, I just picked cyan for the sake of consistency. Something like `{c[Name]{n` could possibly be good as well.

A possible issue could be the ability to see who's online, if they're using `watch/hide` (`subscriptions.online()` might take care of that?) or using it to find out who belongs to what org, if the members are secret for some reason (i.e. Shadow Court), though I'm not sure if the those are that problematic.

### DIFF
--- a/commands/base_commands/general.py
+++ b/commands/base_commands/general.py
@@ -67,6 +67,7 @@ class CmdGameSettings(ArxPlayerCommand):
         @settings/name_color <color string>
         @settings/verbose_where
         @settings/emit_label
+        @settings/highlight_all_mentions
 
     Switches: /brief suppresses room descs when moving through rooms.
     /posebreak adds a newline between poses from characters.
@@ -87,6 +88,7 @@ class CmdGameSettings(ArxPlayerCommand):
     /verbose_where shows any roomtitle information that someone has set
         about their current activities, when using the +where command.
     /emit_label will prefix each emit with its author.
+    /highlight_all_mentions enables highlighting for all mentions in channels.
     """
     key = "@settings"
     locks = "cmd:all()"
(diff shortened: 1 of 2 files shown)


TehomCD:
So I like the idea here, but I think what we might want to do is send a formatted copy to each receiver based on their name matching, rather than have a single message that highlights to everyone if there's a match found. You could do that by calling distribute_message with a new TempMsg using the return value of format_mentions if there's a match for the receiver rather than change the temp message itself.

Thinking about it, it might be better to override distribute message and have some method call there for formatting the message on a per-receiver basis right before it's sent out. Eventually might be worth subclassing TempMsg to hold a lot of shared logic around message transformations on a per-receiver basis, to handle things like languages at a higher level than the .msg method in characters/accounts, which has proven difficult to maintain.

LeBulldoge:
It might still better to highlight the mentions for everyone, as it's useful for others as well, because if there are several ongoing conversations, you might not want to read into messages that aren't directed at you, or someone else you're talking to.

TehomCD:
I think it's a little counter-intuitive to highlight for everyone when anyone is mentioned. We could do `@all` or `@everyone` or something for announcements intended for everyone's notice, but I think it'd just add to noise if any time anyone was mentioned everyone saw the highlight, and the point of something like this is to draw notice when it's something you want to respond to. So, all things considered, I think that I really would prefer if the message is modified on a per-receiver basis.

LeBulldoge:
I think I see what you mean. I'll look into it, see what I can figure out.

LeBulldoge:
On a side note, is there a batch command file somewhere to build an Arx-specific world for testing purposes or something along those lines?

LeBulldoge:
Alright, should be good for review now. I tried looking into subclassing TempMsg as you suggested, but it's a bit outside my league for now, so I decided to go with the easier path.

Tanookami (approved):
Thank you, this looks really good.

TehomCD (approved):

TehomCD:
Thanks a ton for the PR! Unfortunately, no, we don't currently have any sort of quick-start batch script or the like for initializing a database. It's something I definitely want to add at some point.

PR #294, opened 2020-04-27
Channel mentions hotfix

LeBulldoge:
Fixed keeping the same message object over a set of multiple users. Made the algorithm case insensitive.


TehomCD (approved):
yay for case insensitivity

PR #309, opened 2020-06-16
@check/retainer

### LINKED ISSUE #292: @check/retainer - A Discussion
A means of checking retainer skills that are validated by the game.  Currently, @gmcheck is used in order to simulate retainers using purchased skills, but it is given values by the user, rather than the retainer's data.  In other words, "just trust me they have this".

@check/retainer[/flub] \<id>|<@check string here>

Commands that limit retainers to specific abilities rather than rolls in general.  Examples:
* +heal/retainer to allow a retainer to do a healing roll (using that specific stat-skill combo only)
* +work/retainer where that retainer can only roll for an Economic/Military/Social work roll depending on whether it's an Assistant/Champion/Spy respectively (possibly for reduced AP spending or a set amount per week).

**Additional context**
This is to start discussion on retainers using die rolls beyond the scope of the investigation_assistant ability.  As I started implementing this feature, the following questions came to mind:

1. Should retainers be able to roll crits?
2. What cap should they have, if any?  *(I rather think there should be one, myself, not unlike what investigation_assistant does currently; retainers should be for support, **not** to upstage characters, even if the retainer has a 6/legendary skill).*
3. Should their stats and skills be usable (roll eligible) in other systems?  *(Actions and shardhavens both come to mind.)*

dvoraen:
Adds /retainer switch to @check, allowing uses of retainer die rolls.

- Enables game-side checking of retainer stats, rather than using @gmcheck and "just trust me on this" for stat values.  (NOTE: They should not be able to roll crits; this may need double-checked.)
- Increases value of training retainers, because they can supplement PC skills.
- Initially supports only retainer ID rather than name; that's in the works for later.

Closes #292


TehomCD:
Merge conflicts, so it looks like you need to update your branch a bit. I don't recall changing rolling anytime recently, so that's kind of weird.

commands/base_commands/rolling.py
```diff
@@ -68,65 +69,143 @@ class CmdDiceCheck(ArxCommand):
...
+            raise ValueError("There must be one unique match for a character stat. Please check spelling and try again.")
         # get unique string that matches stat
         stat = matches[0]
+
         if skill:
             matches = stats_and_skills.get_partial_match(skill, "skill")
             if not matches:
                 # check for a skill not in the normal valid list
-                if skill in caller.db.skills:
+                if skill in self.caller.db.skills:
```
TehomCD: `.db.skills` is deprecated. If you could, please use the `traitshandler` for checking skills

world/roll.py
```diff
@@ -60,20 +62,31 @@ def __init__(self, caller=None, stat=None, skill=None, difficulty=15, stat_list=
             stat_list = [ob.lower() for ob in stat_list]
             # look up each stat from supplied caller, adds to stats dict
             for somestat in stat_list:
-                self.stats[somestat] += self.character.attributes.get(somestat, 0)
+                if self.retainer is None:
+                    self.stats[somestat] += self.character.attributes.get(somestat, 0)
+                else:
+                    self.stats[somestat] += self.retainer.dbobj.attributes.get(somestat, 0)
```
TehomCD: accessing stats via `attributes` is also deprecated, please get via traitshandler

world/roll.py
```diff
@@ -60,20 +62,31 @@ def __init__(self, caller=None, stat=None, skill=None, difficulty=15, stat_list=
...
             # add individual skill to the list
             if skill and skill not in skill_list:
                 skill_list.append(skill)
             skill_list = [ob.lower() for ob in skill_list]
             # grabs the caller's skills or makes blank dict
-            skills = caller.db.skills or {}
+            if self.retainer is None:
+                skills = caller.db.skills or {}
+            else:
+                skills = self.retainer.dbobj.db.skills or {}
```
TehomCD: `.db.skills` deprecated

TehomCD (changes requested):
I think the merge conflicts are from the conversion to traitshandler and deprecation of evennia Attributes for storing skills/stats. Pretty important you update to have those changes

dvoraen:
Yeah, that was the cause.  I'll have to research why the changes didn't come down, as I was getting the Evennia updates fine (so far as I can tell).  I believe I've resolved all the merge conflicts and deprecated property calls.

TehomCD (approved):
This is great, thanks for the PR!

PR #312, opened 2020-06-20
TT Color Highlighting

### LINKED ISSUE #308: Table Talk Highlight
Socializing and events are not everyone's cup of tea, sometimes they are intimidating by the sheer number of attendees, and sometimes people decide to brave them. Most often there are sufficient places to join, where players can have a tete-a-tete, or a smaller group conversation, separate from the whole. We have tools at our disposal to alleviate some issues had with large scenes, such as the ability to highlight our names, highlights to sayings/quotes, both serving to help grab our attention to ensure emotes are not lost when directed towards us. However, even with these tools, sometimes table talk can get lost in the room emit spam, making it difficult to keep up or follow along with conversation being had at your place.

The solution I'd like to suggest is another settings option that allows players to highlight the place name when talking or emoting at one. While it is something that can be remedied by those savvy with their gaming client and know how to do it, there are those to be considered who don't use an alternative client, but use the web one. 

I think that having the option to highlight a place would alleviate some concern, stress, and/or anxiety had by some players about not being able to keep up, or missing posts, and would ensure that they could be easily seen and referred back to when the event is particularly lively.  I believe that it will make table talk posts easier to spot, and perhaps while helping players, may also encourage further participation with less worry of having to find and search player posts amid ones made to the room at large.

dvoraen:
Adds both optional place name coloring (can turn on/off) as well as which color to label the place name.
Also snuck in a bugfix for @check/retainer because the first thing I did on Live was a syntax error which caused a ListError to be raised.   OOPS.

Enables players using the web client to have highlighted place names.

Closes #308


TehomCD (approved):
LGTM. Thanks!

PR #315, opened 2020-06-23
Place Name Highlight Redux

dvoraen:
Fixing a very major oversight in my feature.  I completely did it wrong.

This second attempt includes a refactor to how TT sends messages (other players were seeing the wrong place colors).  It's now similar to how the Roll class builds a message and then the message is sent out.  I made gratuitous use of str.format() which will allow the messages to be more easily changed if necessary, as well as future possible additions.

Individualizes the highlighted place color as it should have been.

There's an oddity with PC name coloring with respect to the emit_label.  If @settings/name_color is set, the emit_label will not be cyan (unless you set a similar color) when it displays YOUR name.  Other players are cyan.  This is the case on Live as well, but I mention it in passing because the {w[{cNAME{w] lines are somehow being overriden.


TehomCD (approved):
Looks alright to me. In retrospect having two separate attributes for a color highlighting doesn't seem necessary, since `place_color` being defined could be used as a boolean for whether to color things at all.  I can combine those when I get around to redoing settings in the gradual purging of Evennia Attributes. 

I've slightly moved away from enums for constants, but that's not a big deal either. You might consider playing around with f-strings some, I find them much easier to work with than `.format()`.

Anyhoo, thanks for the changes!

PR #322, opened 2020-08-12
Remove 'Favor' from bonus commands

Tanookami:
Removes "favor" from the pool of possible RP commands that grant XP each week. Adds a reminder beside the disabled command to re-add it later, after it's been restored.
Players were encouraged to use a disabled command and this blocked bonus XP.


TehomCD (approved):
Awesome, thanks!

PR #328, opened 2020-10-04
Add new rolling command

TehomCD:
Adds the new stat_check lookup models and the new check command. Old check command is deprecated, but keeps the command key for now. new one is under 'newcheck', will be changed eventually.

Roller for the upcoming new GMing system.


ApostateCD (changes requested):
Make new check the default

ApostateCD (approved):

PR #332, opened 2020-10-04
Blacken codebase

TehomCD:
This runs black on the entire codebase. I'm going to be experimenting with github actions, so I added a workflow that Black's repo suggested - I'm unsure if it runs black by default on all PRs or if I'll need to add that step manually.

Black should be unable to change the functioning of any files: the bytecode generated by it is guaranteed to be the same. So though this is 50k lines and 483 files, it should be safe.


Tanookami (approved):
This is a PR with 485 files changed. 485. 
We're just going to assume it is fine. This is fine. My review ends with a haiku:
```
Your PR is huge
four hundred eighty five files
I love you but no
```

TehomCD:
lol, I love you so much

PR #344, opened 2020-10-09
Private check

### LINKED ISSUE #340: Private checks for new check system
People used private checks a lot with the old system (either targeting just themselves or targeting a GM). It would be great to see this functionality in the new check system, too!

dvoraen:
Adds private-messaging feature to new @check command.

Allows players to privately send rolls to (non-staff*) GMs.  The reason is mostly detailed in #340.

Closes #340 

I added some redundancy checks to reduce the chances of players spamming either staff and/or other players with one command.  Everyone (command user, recipients, staff) "should" only see one message per roll, regardless of who uses the command and how, barring malicious input. :)


world/stat_checks/check_maker.py
```diff
@@ -20,9 +20,15 @@ def check_rolls_tied(roll1, roll2, tie_value=TIE_THRESHOLD):
 
 class SimpleRoll:
     def __init__(
-        self, character=None, stat=None, skill=None, rating: DifficultyRating = None
+        self,
+        character=None,
+        stat=None,
+        skill=None,
+        rating: DifficultyRating = None,
+        receivers: list = None,
```
TehomCD, line 28: typehint of list with a default of None probably raises linting errors because you're not saying it can be empty. The way to do that is with the `Union` value from the typing library, which also has its own container types so that you can specify the values inside them. So it'd be something like:

`receivers: Union[List, None] = None`

With List, you could also specify `List[str]` to show that it'd be a list of names. More on that later though...

world/stat_checks/check_maker.py
```diff
@@ -20,9 +20,15 @@ def check_rolls_tied(roll1, roll2, tie_value=TIE_THRESHOLD):
...
     ):
         self.character = character
+        self.receivers = receivers
```
TehomCD: For receiving an iterable argument that can be None it's safer to set it to be an empty list/iterable so that it is still iterable and won't raise type errors if you try to iterate on it. So I'd prefer:
`self.receivers = receivers or []` for example so that it would be an empty list rather than None when not specified.

world/stat_checks/check_maker.py
```diff
@@ -88,6 +94,68 @@ def announce_to_room(self):
...
+        """
+        self_list = (
+            "me",
+            "self",
+            str(self.character).lower(),
+            str(self.character.key).lower(),
+        )
+
+        # Adjust receiver list; this is who actually will receive it besides staff and the caller.
+        receiver_list = self.receivers.copy()
```
TehomCD: if `self.receivers` is None here this will raise an AttributeError. Right now that won't happen since `announce_to_players` is only called in a case that also populates receivers, but it'd be nice to prevent that from happening if this was extended/changed in the future.

world/stat_checks/check_maker.py
```diff
@@ -88,6 +94,68 @@ def announce_to_room(self):
...
+        for name in self.receivers:
+            receiver = self.character.search(name.strip(), use_nicks=True)
```
TehomCD: While we can do the search here, in general I'd prefer if we did the search in the command and passed the actual objects instead. Two reasons for that: I think from a separation of concerns standpoint the command should be responsible for interpreting player input, so anything related to parsing player input is better handled there, and also so that we can more easily handle cases like rolls where receivers are in different rooms. They might be specified by some other mechanism rather than just by name: for example, it could be everyone on an action together. When generating a roll for that, we'd want to pass along the receivers based on the action, rather than based on a command - making it based purely on names here and executing a search prevents that from working.

world/stat_checks/check_maker.py
```diff
@@ -88,6 +94,68 @@ def announce_to_room(self):
...
+        receiver_suffix = "(Shared with: %s)" % (", ").join(receiver_list)
+        private_msg = f"|w[Private Roll]|n {self.roll_message} {receiver_suffix}"
+
+        # Always sent to yourself.
+        self.character.msg(private_msg, options={"roll": True})
+
+        # Send result message to all staff in location.
+        staff_list = [
+            gm
+            for gm in self.character.location.contents
```
TehomCD, line 153: Similarly, `self.character.location` is not guaranteed to exist. Characters who are offline have no location, and when porting this to actions we'd want to only have this check if they have a location before checking its contents

TehomCD (changes requested):
Just a few minor changes I'd like made so that this can be portable to actions and other checks eventually.

TehomCD:
I fixed tests and updated your branch here, so before making the changes I asked for you'll probably wanna do a `git pull` locally to grab stuff that was merged in

world/stat_checks/check_maker.py
```diff
@@ -25,10 +25,10 @@ def __init__(
         stat=None,
         skill=None,
         rating: DifficultyRating = None,
-        receivers: list = None,
+        receivers: list = [],
```
TehomCD: So there's a real gotcha here in python for mutable default arguments. Because the signature of a function is in the class, it's in the global/module scope and is allocated at import time. What this means is that there's a list as the default argument - and that list is shared by everyone who ever calls that function. So if it changes in any way, it'll affect other people who call the function.

For example, if someone were to have the default argument for receivers and then append something to receivers elsewhere, the next call that has the default argument would have the modified version. It can be a source of super confusing errors. tldr - use `None` like you did before and set the assigned class attribute to an empty list like you do in line 31

dvoraen: I appreciate the sage advice.  I need to go find a book on Python that matches some of my C++ library regarding dos and don'ts.

world/stat_checks/check_maker.py
```diff
@@ -88,6 +94,82 @@ def announce_to_room(self):
...
+            # Also checks for duplicate receivers to cross them off too so malicious users
+            # can't put "=Name,Name,Name" and spam the other party with the result.
+            if (
+                receiver.check_permstring("Builders")
+                or receiver.name.lower() in self_list
+                or receiver_list.count(receiver) > 1
+            ):
+                receiver_list.remove(receiver)
+            else:
+                receiver_names.append(receiver.name)
```
TehomCD: while this works there's simpler approaches. python has a `set()` built-in type that is unordered but has no duplicates. So an easy way to do this would be something like:
```
receiver_list = [ob for ob in set(self.receivers) if not ob.check_permstring("Builders") and ob.name.lower() not in self_list]
receiver_names = [ob.name for ob in receiver_list]
```

world/stat_checks/check_commands.py
```diff
@@ -50,6 +54,22 @@ def do_normal_check(self):
             self.caller, stat=stat, skill=skill, rating=rating
         )
 
+    def do_private_check(self):
+        receiver_names = [player for player in self.rhs.split(",")]
+        receiver_list = []
```
TehomCD, line 58: just as an fyi there's already an attribute in classes that does this: `self.rhslist` is the arguments on the right hand side of the `=` that are comma-separated

TehomCD (dismissed):
Looks good! There's a few nitpicks if you feel like tweaking it, but nothing that I'd block for. Thanks a ton! If you want to make further changes, feel free and I'll review again, otherwise I'll merge it in

TehomCD (approved):

PR #345, opened 2020-10-10
Filter unviewable objects from character search

### LINKED ISSUE #326: Prune search results by visibility
Either using view lock directly or a new method, prune the list of objects passed from contents to search in the character search method based on whether the character can see them. Bypass for `is_staff`.

Tanookami:
Adds a character search-result filter on objects they don't have view-access for, similar to the filter that prevents searching for a masked character by their real name.
If someone can't see it, they should not be able to search for it. While a more robust system will allow special forms of sight or skillchecks to reveal the hidden, this will help prevent awkward spoilers until then.
Closes #326


TehomCD (approved):
Perfect! Thank you so much <3

PR #355, opened 2020-10-31
Updated Prayer Display

ApostateCD:


TehomCD (approved):
🎵 Take my hand, We'll make it I swear 🎵

PR #356, opened 2020-11-01
Private roll display adjustment

dvoraen:
This modifies how names are displayed for making private rolls with new @check.  I also cleaned up the code a little and took out some redundancies, including one of my meaningless tests.

OLD: 
\- Would display self-only if you're sending it to yourself and staff (because staff can see it anyway).

NEW:
\- Names are sorted() and grouped by PCs and staff, in that order.
\- Will display staff names if specified to receive a private roll.  Staff names will be listed last in the list as well as have a highlight.  This is done to make it easier to find their names.
\- Self-only will display only if you're sending it to yourself and no one else (but staff will still see it).

Player feedback and uncertainty about whether GMs were getting private rolls or not.

### DIFF
--- a/world/stat_checks/check_maker.py
+++ b/world/stat_checks/check_maker.py
@@ -106,37 +106,32 @@ class SimpleRoll:
             str(self.character.key).lower(),
         )
 
-        # If we fail to make a copy (e.g. self.receivers is None),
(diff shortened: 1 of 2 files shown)


TehomCD (approved):
Looks good, thanks!

PR #363, opened 2020-12-27
Caravel

ApostateCD:


TehomCD (changes requested):
There's a learning experience here with something that's incomplete.

Django migrations work by django running through them all and trying to build up what it thinks the state of your models is based on the operations inside them. If the state it gets from the migrations doesn't match what it sees in your models, it'll suggest you need to generate a migration to fix it. I'll never actually inspect the database itself, beyond a single table that tracks which migrations are applied: the models in your apps and the migrations files are all it has to go on.

Why does this matter? Because one of the things it tracks as part of the model's state is the choices field which you modified, even though that's never stored in the database: it only lives in the migrations and the model definitions.

So when you made that change, django will think it has to generate a new migration to fix its state, even though it's not a change to the database schema at all: it's something that only exists in its model definitions and state generated by migrations. While you can just make a new migration to fix that, it's not the best solution: migrations have a non-trivial overhead when running tests, and applying even a migration like that would be unwise without taking the game down.

Instead, if you change the existing migration that added that `unit_type` field, you can add the option there, and django will see that now the state and the models are the same. Since the migration is already applied on our dev/prod environments, nothing has to be done. Also since the change never hits the database, the migration already being applied doesn't matter.

tldr - change the migration to add the caravel choice to the [unit_type field](https://github.com/Arx-Game/arxcode/blob/stable_orphan/world/dominion/migrations/0001_squashed_dominion.py#L2472).

edit: add link

TehomCD (approved):

PR #368, opened 2020-12-29
Notifier w/ private check changes

dvoraen:
Notifier is a class hierarchy that encapsulates sending in-game messages to subsets of players based on specific notifier behavior.  Multiple notifiers can be used in tandem to generate varied behavior in how notifications are sent out.  See below for more information.

Notifiers initially support filtering receivers to non-GM players, GM players, and staff in any combination.

Initial notifiers are as follows:
- Notifier: base class meant to be derived from; contains utility and default code that can (and sometimes must) be overriden by subclasses
- RoomNotifier: notifier that sends notice to a room
- ListNotifier: notifier that sends notice only to the provided list of names
- SelfListNotifier: notifier that sends notice to caller and provided list of names

Future expansion ideas (which may or may not see fruition):
- ShardhavenNotifier: notify all designated characters in a given shardhaven
- DomainNotifier: notify all designated characters in a given domain
- OrganizationNotifier: notify all designated characters in a given organization
- GlobalNotifier: notify all designated characters in the game (possible expansion to @wall's capabilities)

This is the start to providing a relatively easy way to send out mass notifications without resorting to list comprehensions and other Python constructs in order to get a message from command A to recipients A-E.

The private @check changes utilize two notifiers: one RoomNotifier which notifies all staff in a room (staff see all private rolls) and one SelfListNotifier which notifies the caller and the designated players provided by the caller (in addition to staff).

The private @check behavior should be mostly the same: player designated the list of receivers and the player, those receivers, and staff, all get notified of the roll.  The player does see that staff receives the roll is one change I put in.

Also, I'm really scared to submit this.  I have that niggling doubt that I forgot something, or something else is broken, despite testing.


server/utils/notifier.py
```diff
@@ -0,0 +1,177 @@
...
+        self,
+        caller,
+        **to_flags,
+    ):
+        self.caller = caller
+        self.to_flags = to_flags
+
+        self.receiver_set = set()
+
+    def notify(self, msg: str, options: dict = None):
```
TehomCD: Not that I care much, so this is just as an fyi, but having a type annotation with a default value of `None` will raise linting errors with more strict linters. The way around that is with the typing library, having a `Union[Dict, None]` as the type. But codacy didn't care so it's whatever

server/utils/notifier.py
```diff
@@ -0,0 +1,177 @@
...
+class RoomNotifier(Notifier):
+    """
+    Notifier for sending to everyone in a room, filtered by
+    the to_ flags.
+    """
+
+    def __init__(
+        self,
+        caller,
+        room: ArxRoom,
```
TehomCD: I'd rather not have an import of a typeclass in a utility file because it'd be really easy to have circular imports occur. You can do a lazy reference of "ArxRoom" most likely, which a lot of linters will fail to resolve, or just skip it entirely. I just really hate any imports of other project modules inside a utility module if you can avoid it

TehomCD (approved):
Looks awesome, thanks!

PR #369, opened 2021-01-03
Porting GMCheck

dvoraen:
This is to port @gmcheck over to the new check system.

Self-explanatory: New system, port the old stuff!

I also added an optional feature to it: =\<NPC name\>.  Players can designate an NPC name to better track which of their mooks rolled what stats/skills.  In the future, I'd love it if the jinja2 templates would get updated to accommodate for npc_name being present, to allow rolls to say "NPC was successful/failed/catastrophically fails/etc." if a name was provided (PC name by default), but this is a feature request more than an implementation detail.

IMPORTANT NOTE: I largely retained the old gmcheck's functionality, specifically the point about critical rolls not occurring without the switch.  Given how crits and botches work, GMcheck rolls don't crit OR botch unless /crit or /flub is used.


dvoraen:
I'm pretty sure Codacy is going to fail on one point, because I don't have that set up yet to check for me.

TehomCD (approved):
Looks good, thanks!

PR #371, opened 2021-01-08
Porting @check/retainer

### LINKED ISSUE #63: @check/retainer
Retainers have stats and skills, which is cool! But there's currently no way to do a `@check` with their stuff other than `@gmcheck`, where you have to put the numbers in manually. Which is both a hassle and, well, people can kind of say their retainer stats are -- whatever they want to say they are. So you don't have the same sort of assurances.

`@check/retainer`! I'm thinking of the syntax `@check/retainer <retainer ID#>/<stat>[+<skill>][ at <difficulty number>][=receivers]`. So as close as possible to the original `@check` syntax, just with that retainer ID thrown in first. The idea of course being to do a dice roll of the given retainer stat+skill and display it like a PC:

`<DICESTRING> RETAINER NAME checked dexterity + athletics at difficulty 15, rolling 52 higher.`

dvoraen:
Porting over @check/retainer

Porting over my baby, but the motivation has to do with #63's reasoning.

Closes #63 

This pull request also has a +request in game for it (see ticket 25887).  Please review that before this pull request.


TehomCD:
I read your request, but I think adding unused context variables isn't the best solution to the problem. In general, you should only have logic inside templates if there's a compelling reason to do so and keep more logic in what's populating the template - usually views, but in this case commands. In the example you provided, of changing the roller to {{npc_name or retainer or character}}, that doesn't make sense to me over changing the object being passed that will appear as character. Right now, the only thing character is used for in the template is just a name of the roller, so it makes a lot more sense to me to substitute that out for whatever name you want when populating the template.

world/stat_checks/check_maker.py
```diff
@@ -388,6 +388,123 @@ def _get_fail_rolls(self):
...
+        botch = None
+        if self.natural_roll_type:
+            if self.natural_roll_type.is_crit:
+                crit = self.natural_roll_type
+            else:
+                botch = self.natural_roll_type
+
+        return {
+            "character": self.character,
+            "npc_name": self.npc_name,
```
TehomCD: changing "character" from `self.character` to `self.npc_name` removes the need to change the template

world/stat_checks/check_maker.py
```diff
@@ -388,6 +388,123 @@ def _get_fail_rolls(self):
...
+            elif self.retainer.name.count("-") >= 1:
+                short_name = self.retainer.name.split("-", 1)
+                short_name = short_name[0].strip()
+            else:
+                short_name = self.retainer.name
+        except (ValueError, IndexError):
+            short_name = self.retainer.name
+
+        return {
+            "retainer": short_name,
```
TehomCD: same here - replace "retainer" with "character", and there's no need to change the template. Ideally, from the template's perspective, all it should know is that it's getting an object to represent who made the roll. Unless we need to present information differently based on their nature, I'd keep it decoupled

TehomCD (changes requested):
As per the comments, I'd prefer 'character' be used to represent retainers, PCs, or spoofed rolls in the context and hide details from the template it doesn't need to know

world/stat_checks/check_maker.py
dvoraen: I didn't think of that.  I'll have to look at that tomorrow when I'm more lucid.  It'll require a little TLC (but nothing unmanageable) because @gmcheck may or may not have self.npc_name defined; it depends on the input which one would go to the template.  (Writing this as a note to myself, too.)

TehomCD (approved):

PR #391, opened 2021-05-12
@adjust multi-char support

dvoraen:
This upgrades the @adjust staff command to support multiple characters on the command line.

Apostate said it would be really swell if I could do that, so I did.

Areas needing fine-toothed comb:
- Usage of the transaction.atomic() context manager and its enclosing try/except block.  I'm not 100% certain I have that correct logically.  I'm particularly concerned about the "no one was adjusted" exception being an accurate error to raise, even though I don't foresee it actually happening in practice.
- I am completely reliant upon my test cases for this command upgrade. I wasn't able to really test it in action on my private server due to technical issues on my end that I haven't solved.  Logically they're succeeding, which is heartening, but I'm hoping I didn't forget anything obvious.

### DIFF
--- a/commands/base_commands/staff_commands.py
+++ b/commands/base_commands/staff_commands.py
@@ -3,9 +3,10 @@
 Admin commands
 
 """
-from typing import Tuple, Optional
+from typing import Tuple, Optional, List
 
 from django.conf import settings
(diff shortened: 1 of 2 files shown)


TehomCD (approved):
Looks great, thanks!

PR #396, opened 2021-06-27
Changed emit label to differentiate from channel labels

LeBulldoge:
Changed emit labels from `[Name]` to `(Name)`, as the former currently matches channel labels, which breaks triggers. Since you're allowed to pose within channels as well, this is virtually impossible to avoid via regex, unless you were to manually specify every channel you want to fall under a trigger.
Working triggers maintain the player's sanity level.
Picked parentheses to maintain some similarity to the original label, but could be changed to something that fits better, if there are any alternatives.


TehomCD (approved):
Looks good, thanks!

Issue #357, opened 2020-11-15
Let players close their own requests


rozzingit:
This is one that actually gets asked for a fair amount on Info. But basically: a `request/close` or `request/cancel` command on the player side so that we can close our own requests when they become obsolete/we no longer need them. Which should also help to reduce your list of open jobs!

TehomCD:
Closed by #397

PR #397, opened 2021-07-01
Allow players to resolve their own tickets

LeBulldoge:
Added a `/close` switch to `request`, letting players close their own tickets and supply a resolution.
Should help reduce ticket bloat
Closing #357


LeBulldoge:
Also fixed the failing test in the stable branch, which was caused by the `rating` field being no longer required in `ClueCreateForm`, though not sure if the fix is correct, the test might be outdated.

And I don't really agree with/understand the remaining lint checks, one of them is even complaining about a migration file. :o

TehomCD:
Oh the github action for black can't be pinned to a version anymore due to some upstream issue, so if someone wisely decided to change indentation in a new version or whatever everything explodes. I'll see about fixing that at some point today

LeBulldoge:
Alrighty, thanks. I believe everything should be ironed out now.

TehomCD (approved):
Looks great, thanks!

PR #410, opened 2021-10-06
Fix a bug in check/vs

### LINKED ISSUE #351: @check/vs Exception
It's a bug!

If one uses @check/vs immediately after the server comes up or is reloaded, an exception emerges.  If @check is used to do a roll before the /vs command, then all is well in the land and @check/vs will work accordingly.

**Additional context**
check_commands.py line 142 is the offender
`        rating = DifficultyRating.get_all_cached_instances()[0]`

Tanookami:
a variable was using the wrong method, one that didn't check for the existence of a cache.
We fix bugs. :bug:
closes #351


TehomCD (approved):
Awesome! Thank you 💖

PR #411, opened 2021-10-10
Fix clues/search

### LINKED ISSUE #353: @clue/search includes context notes
When using @clue/search (as opposed to @clue/search/tag), the search function will search the context notes attached to @clues in addition to the @clues themselves.

E.g. on Katarina I might type "@clue/search Eurus" and mixed in with the results will be:

_[ The Colonel's Other Butt ] (666 rating) Tags: The Other Metallics, The Double Secret Archfiends, butts
As it turns out, the colonel had ANOTHER butt, which he kept hidden, and in which he hid other secrets besides.
10/27/20 22:56:48 This clue was shared with you by Ember, who noted: Ember shares this clue with the Princess of **Eurus** in flagrant violation of alt interaction rules._

I don't know if it would be possible to limit what @clue/search searches to omit those context notes without gutting the code or making more work than is feasible right now, but if it IS possible, I think it would make @clue/search that much more useful.

Tanookami:
Changes clues/search command to exclude text in clue discoveries. Adds the clues/search/all switch to do that instead.
Makes searching for clues more specific.
closes #353 Happy Hacktoberfest 🎃


TehomCD (approved):
Awesome work, thank you! :shipit:

PR #415, opened 2021-10-23
Removing mask restrictions and added logging

ApostateCD:
This removes mask degradation and movement restrictions, and added in a log file to track who is wearing what masks.


TehomCD (approved):
Very nice change! :shipit:

PR #419, opened 2021-10-30
Room pet controls

### LINKED ISSUE #302: No Pets Allowed
People bring pets into all sorts of places they shouldn't. As in: it shouldn't be ICly feasible, but there's nothing stopping them OOCly/codedly, so they just assume it's fine.

Code you can set on a room to basically block the entrance of animal and small animal retainer types. Probably best suited as a manageroom toggle (manageroom/pets?) so that people can have control over it for their own room/shops/builds, although tbh I think the default should be No Pets Allowed and people can switch it if they want. Definitely should be set on all the shrines and all the various inns in the different boroughs, along with the Assembly of Peers. Those are the first ones that come to mind, anyways. (Maybe also the Palace?)

Anyways the idea is that if you had animal retainers with you they'd either auto-dismiss when you entered with a pemit about pets not being allowed, or else actually block entrance until the player dismisses them manually. The later might be more effective at communicating the message, tbh; it'd be easy to miss an auto-dismissal even with a message, because it'd immediately scroll up with the room desc. (Maybe include an OOC bit in the message that reminds them that the command is 'guards/dismiss <retainer>' or whatnot.)

Asking people OOC every time to Please Not

Anyways @ApostateCD already said he was gonna do this but I was already halfway through the request, so.

Tanookami:
Adds a switch to the `+manageroom` command. Togglepets will control whether characters can enter while their pets (animal retainers) are following. Providing comma-separated character names will toggle each character on/off an "allowlist" for exemptions to the room's pet ban. When the room is toggled to allow all pets again, that list is cleared. And typing "allowlist" instead of names returns a message about the room & allowlist status. Even on the allowlist, disguised individuals with animals will be blocked to preserve anonymity.
This should preserve gravitas in locations where animals would seem inappropriate. I wanted the command to be simple & lightweight, easy to change per event, instead of a complex permanent system of permissions.
Closes #302 . Hopefully. Happy Hacktoberfest 🎃


TehomCD (approved):
🎃 this is so awesome 🎃 

I love it! People have been asking for this for ages

PR #422, opened 2021-10-31
Moved OC turn off to setconfig command

ApostateCD:
OC creation moved from staffbreak command to setconfig.


TehomCD (approved):
Very nice to decouple this from the staff break. 🎃 :shipit:

PR #426, opened 2021-11-01
Hide masked characters from substrings

TehomCD:
Masked characters can be outted based on substring matches. This ensures that their identities stay hidden, though there could still be some fuzziness if there's overlap between a mask and their key.


Tanookami (approved):
thanks for this!

PR #427, opened 2021-11-02
Fix Togglepets

Tanookami:
Changes properties related to the `+manageroom/togglepets` switch. Compares the location of pet & owner to determine if pets are following. Adds a verbose error for masked characters. And fixes some logic for the allowlist toggle, allowing names to be added instead of replacing the whole list at each use.
Pets are not removed from the list of a character's followers until the person goes somewhere else. After dismissing pets, characters were still not allowed to traverse into a pet-banned room unless they changed rooms first.
✌️ Happy Hackvemberfest?


TehomCD (approved):
Outstanding, thank you! ❤️ 🎃 

:shipit:

PR #430, opened 2021-12-27
fixed the fealty references to make census work

ApostateCD:
Should be a very simple fix that makes the +census command work again, just changed the references.


TehomCD (approved):
LGTM

PR #434, opened 2022-02-08
removed deprecated text from action help

ApostateCD:
Just removed the code block that did the actions based on time stuff, since it's once per episode now.


TehomCD (approved):
LGTM

PR #443, opened 2022-06-16
Update git actions and upgrade black version

dvoraen:
Updates
- set black version to 22.3.0, including reformatting the project files.  10 files were changed, which included some of the migrations for web/helpdesk due to them using Python 2 syntax such as u-strings and print without parentheses.  black 22.3.0 includes some formatting changes, including:
  - stripping out u-string prefixes (u"" becomes "")
  - compressing together ** operator usage if the left and right side are "simple" expressions, including nested attributes (function calls and the like retain a space on the left and the right operator)
- updated black workflow action to use python 3.8 and black 22.3.0
- requirements.txt updated to accommodate for library changes
  - pinned markupsafe to <2.1 as v2.1 of markupsafe removed code that our version of Jinja2 was importing
  - updated black to 22.3.0

- also specified Python 3.8 in the Dockerfile while I was at it

Because I care. <3

Pretend I'm dependabot, but not really.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -7,6 +7,7 @@ dnspython==1.16.0
 graphviz==0.13.2
 humanize==0.5.1
 Jinja2==2.11.3
+markupsafe<2.1
 pillow==8.3.2
 python-decouple==3.3
 requests==2.22.0
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,4 +1,4 @@
-black==21.4b2
+black==22.3.0
 cloudinary==1.19.1
 django-bootstrap-form==3.4
 django-crispy-forms==1.8.1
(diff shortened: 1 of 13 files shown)


requirements.txt
```diff
@@ -7,6 +7,7 @@ dnspython==1.16.0
 graphviz==0.13.2
 humanize==0.5.1
 Jinja2==2.11.3
+markupsafe<2.1 # v2.1 removes code used by jinja2 v2.11.3
```
TehomCD, line 10: Where's this dependency used?

dvoraen: Jinja2 uses markupsafe as part of its library.  Markupsafe 2.1 removes soft_unicode which is used by our version of Jinja2 so the test suite will fail if we don't pin markupsafe to a version that still retains the soft_unicode function.

BUT.  Jinja2 (at least, v2.11.3) doesn't seem to pin their dependencies to a "highest compatible version" (their changelog mentions >= in their dependencies) so the test suite will download the latest version of markupsafe as part of installing jinja2, which then creates ImportErrors.  This can be seen best in one of my action logs:

https://github.com/dvoraen/arxcode/runs/6925008225?check_suite_focus=true

This is the issue as reported by others, if you want a reference: https://github.com/pallets/jinja/issues/1585

PR #444, opened 2022-07-11
Created a +coinflip command.

ApostateCD:
Lets characters flip a coin.

### DIFF
--- a/commands/base_commands/social.py
+++ b/commands/base_commands/social.py
@@ -3030,6 +3030,86 @@ class CmdThink(ArxCommand):
         caller.msg("You think: %s" % self.args)
 
 
+class CmdCoinFlip(ArxCommand):
+    """
+    Flips a coin, calling heads or tails or without a call.
+
+    Usage:
+        +coinflip <heads or tails>
+            or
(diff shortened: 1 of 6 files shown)


commands/base_commands/social.py
```diff
@@ -3030,6 +3030,86 @@ def func(self):
...
+            if flip == "heads":
+                caller.msg(
+                    "You flip a coin calling tails, but it lands on heads. You lose. Aw."
+                )
+                txt = (
+                    self.caller.name
+                    + " flips a coin calling tails, but it lands on heads. They lose. Aw."
+                )
+                caller.location.msg_contents(txt, exclude=caller)
+                return
```
TehomCD, line 3108: You can DRY this out substantially by checking if args `is in ("heads", "tails")` and then checking if `flip == args` for a win, loss otherwise

TehomCD: Also you probably want to cast args to lowercase to make it case insensitive with `args = self.args.lower()`

TehomCD (approved):
Approving, will let you refactor later

PR #448, opened 2022-07-19
Added validation to CG with error messages for characteristics

ApostateCD:
Validated inputs so now CG skintone, haircolor, eyecolor, religion check valid inputs.

### DIFF
--- a/commands/base_commands/guest.py
+++ b/commands/base_commands/guest.py
@@ -90,6 +90,41 @@ _valid_fealty_ = (
     "Valardin",
     "Velenosa",
 )
+_valid_religion_ = (
+    "Pantheon",
+    "Shamanism",
+    "Pantheon/Shamanism",
+)
+_valid_hair_color_ = (
+    "Red",
+    "Brown",
+    "Black",
+    "Blonde",
+    "Gray",
+    "White",
+    "Bald",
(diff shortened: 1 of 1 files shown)


commands/base_commands/guest.py
```diff
@@ -90,6 +90,41 @@
...
+    "Olive",
+    "Dusky",
+)
+_valid_eye_color_ = (
+    "Brown",
+    "Blue",
+    "Green",
+    "Hazel",
+    "Gray",
+)
```
TehomCD, line 126: This is an anti-pattern that was used early on and is being continued here - ideally, all these values should live in the database so can be controlled by staff at will. We can hold off on that implementation until CG is reworked, however, because you'd probably want to do it via the web and it's much easier to have a form or serializer use options based on a query than it is to do the same thing here

TehomCD (approved):
Looks okay for patching  up legacy code that we'll move away from, just remember eventually those options should be populated via query rather than using hardcoded constants

commands/base_commands/guest.py
ApostateCD: Yeah that makes sense, we'd make a CG character set on django and then have it pull from that, got it.

PR #449, opened 2022-07-20
Pretty much echoing exactly what you said

ApostateCD:
Yeah as soon as I did it there were concerns about a couple of the choices I had, so there's that anti-pattern pretty much exactly as you thought. Fixed it and then will fix things properly down the line.


TehomCD (approved):
You were attacked with a powerful curse named "Bitter Experience". Anyway, that's part of the charm of django - it has stuff built-in for populating forms with options based on relationships between models. You'll see how to do that when you move this to being a webpage.

PR #454, opened 2022-07-27
Updated requirements.txt to show right pip install

ApostateCD:
For django-sendgrid-v5==1.2.1


TehomCD (approved):
Nice, thanks for changing the hardcoded values to use settings

PR #455, opened 2022-08-01
I changed the wording for @wall because it confused players

ApostateCD:
Wall now says broadcasts rather than shouts. Go team!


TehomCD (approved):
An absolutely critical change, to be sure. LGTM

:shipit:

PR #457, opened 2022-08-02
Census fix, thanks for talking me through it

ApostateCD:


TehomCD (approved):
Eventually may want a 'public' field or something associated with a Fealty and query it as a count through them - this won't show fealties that have 0 players in it, as an edge case

PR #458, opened 2022-10-17
fix render_to_response

Tanookami:
Fixes some import paths of deprecated functions for Django 3.0
one of the necessary steps for upgrading Django
Happy Hacktoberfest 🎃 :shipit:


TehomCD (approved):
Awesome work! We'll probably need to make an epic for all the changes that have to be made for upgrading to django 3.0, but it's definitely a priority because Evennia has already moved to it. Thanks a ton!

PR #459, opened 2022-10-24
Remove static files from source control

Tanookami:
We're getting rid of static files so they aren't tracked. 
Reduces hassle for new servers that would produce a diff from these files which should not be tracked.
Happy hacktoberfest 🎃


TehomCD (approved):
That is one heck of a diff. Thank you <3

PR #461, opened 2022-10-26
Fix relative imports

Tanookami:
Turning relative imports into absolute for future-proofing and stability.
Relative imports are a bad idea in general.
Happy Hacktoberfest 🎃


TehomCD (approved):
Thank you so much. Relative imports are a major barrier to integrating django with sentry when we eventually add that, so this helps a ton for future-proofing

PR #462, opened 2022-10-26
Fix static paths

TehomCD:
After removing the static files from source control, it became apparent that there were a few resources with broke static paths in our templates - namely, css files and a few images. This resolves their paths so they load.


Tanookami (approved):
Thank you for fixing the site. It will look much better fixed than broken! :shipit:

PR #463, opened 2022-10-26
Ignore another fealty graph file

TehomCD:
While playing around with static files I was surprised to see another fealty graph file generated without a file extension. This just adds it to gitignore to prevent it accidentally being committed.


TehomCD:
For reference, the fealty graph is this auto-generated graph that shows up on the world page:
image

Tanookami (approved):
Thank you for catching this.

PR #465, opened 2022-10-30
Add backup_db command

TehomCD:
This adds a command for safely creating a backup of the default sqlite database that Evennia uses. It requires that sqlite3 be installed in the system where you run it (which is also required by the `dbshell` command). While a sqlite database is just a file, copying it while it's under use will generally result in the database becoming malformed and unusable - the backup command is much safer.


Tanookami (approved):
Wow, this is a really great addition! More automation is great. :shipit:

PR #467, opened 2023-02-15
Fixed orgs always being treated as taken an action in an episode by c…

ApostateCD:
…reating new episode foreign key


world/dominion/admin.py
```diff
@@ -690,14 +690,23 @@ def queryset(self, request, queryset):
...
+        "org",
+        "plot",
+        "player_action",
+        "week",
+        "status",
+        "episode",
+    )
     search_fields = ("plot__name", "=dompc__player__username", "=id", "org__name")
     list_filter = (CrisisListFilter, "status")
     raw_id_fields = ("dompc", "gemit", "gm", "plot", "beat", "org")
```
TehomCD: add episode to raw_id_fields

world/dominion/plots/models.py
```diff
@@ -1475,7 +1483,7 @@ def check_action_against_maximum_allowed(self):
         if self.org:
             action_ids = (
                 self.org.actions.exclude(id=self.id)
-                .filter(Q(beat__isnull=True) | Q(beat__episode=episode))
+                .filter(Q(episode=episode))
```
TehomCD: This will remove the previous condition so that they no longer apply at all. If you want them to continue to do so, you can have a logical OR with a `|`. Otherwise you don't need a `Q()` object - it's only for complex queries

world/dominion/plots/models.py
```diff
@@ -1064,6 +1064,14 @@ class PlotAction(AbstractAction):
         related_name="actions",
         on_delete=models.SET_NULL,
     )
+    episode = models.ForeignKey(
+        "character.Episode",
+        db_index=True,
```
TehomCD: `db_index=True` is the default for foreign keys - default args should be omitted

world/dominion/plots/models.py
```diff
@@ -1064,6 +1064,14 @@ class PlotAction(AbstractAction):
...
+        blank=True,
+        null=True,
+        related_name="episode",
```
TehomCD: related_name is referring to the objects that point to the episode, so episode is misleading. It should also be plural, since a FK reflects a many-to-one relationship, with the episode being the "one" side of that. Since this is a plot action, it should probably be `"actions"` or `"plot_actions"`.

commands/base_commands/tests.py
```diff
@@ -559,7 +559,7 @@ def test_cmd_action(self, mock_get_week, mock_inform_staff):
         action_9.save()
         action_9.status = PlotAction.NEEDS_PLAYER
         action_9.save()
-        self.call_cmd("/submit 9", "Org has taken an action.")
+        self.call_cmd("/submit 9", "You have submitted your action.")
```
TehomCD, line 562: Previously this was a test that returned the failure message for an org taking an action. It'd probably be better to attach an episode to it, get the same failure message, then remove the episode from it, and have it pass, so you can test the regression you're fixing

ApostateCD:
So I added episode to raw_ideal fields and removed the default db_index=true, but I ran into a problem with removing that or behavior in the complex query, and I can't tell if it was the fault of that or something else. I noticed on live sometimes changes we made reverting to different database fields, particularly when we changed in django. Here I've had the foreign_key pointing to an episode in an action sometimes vanishing. This seemed much worse when I had the check for a beat, and I wasn't able to pin down why yet.

When trying to cancel an action on my stage environment, we had the ooc_intent field's foreign key's on_delete behavior not match the schema, which had gone to undefined and made deletes fail. I changed that in sqlite to mirror the model, and I checked production- it works fine there, so it was just a local problem. But what I'm not sure is if randomly sometimes losing data with foreign keys vanishing is in any way related, and since I can't consistently reproduce it yet, I don't have an answer there.

But as long as the foreign keys aren't having any kind of dataloss, the checks work fine.

world/dominion/admin.py
```diff
@@ -690,14 +690,23 @@ def queryset(self, request, queryset):
...
+    )
     search_fields = ("plot__name", "=dompc__player__username", "=id", "org__name")
     list_filter = (CrisisListFilter, "status")
-    raw_id_fields = ("dompc", "gemit", "gm", "plot", "beat", "org")
+    raw_id_fields = ("dompc", "gemit", "gm", "plot", "beat", "org", "episode")
     readonly_fields = ("ooc_intent",)
     filter_horizontal = ("search_tags",)
     fieldsets = [
-        (None, {"fields": [("dompc", "topic"), ("org",), ("search_tags",)]}),
+        (None, {"fields": [("dompc", "topic"), ("org",), "episode", ("search_tags",)]}),
```
TehomCD: Can this take a string rather than tuple?

world/dominion/plots/models.py
```diff
@@ -806,7 +806,7 @@ def check_org_busy(self: "PlotAction"):
             if (
                 self.org.actions.exclude(status=PlotAction.DRAFT)
                 .exclude(status=PlotAction.CANCELLED)
-                .filter(Q(beat__isnull=True) | Q(beat__episode=episode))
+                .filter(Q(episode=episode))
```
TehomCD: Q() with a single arg is a no-op, so you can remove that. It can just be filter(episode=episde)

TehomCD (changes requested):
Fix related name

world/dominion/plots/models.py
TehomCD: This still needs fixed

world/dominion/admin.py
ApostateCD: Seemed to work fine but don't see any reason to risk it, so let me change that.

world/dominion/plots/models.py
ApostateCD: Fixed

TehomCD (approved):

PR #468, opened 2023-03-01
Infinite Loop protection in weather

ApostateCD:
I added in a max number of 10 attempts to the while loop, and near as I can tell the script handles a none result for emit. Still not entirely certain how the error first started. I think the most likely culprit is weather was set in game, the script was deleted in django, tried to advance and found it no longer existed and couldn't reference any weather fields. It's hard for me to tell for sure, and checking with Herja to see if I can reproduce it locally.


TehomCD (approved):
Looks ok for now. Probably eventually should have errors related to detecting weather without emits.

PR #469, opened 2023-04-11
Very simple change to add heirloom and legacy tags for items

ApostateCD:
Just alternate tags than plot for GM use and that won't be junk-able.

### DIFF
--- a/world/crafting/junk_handlers.py
+++ b/world/crafting/junk_handlers.py
@@ -39,7 +39,9 @@ class BaseJunkHandler:
     @property
     def is_plot_related(self):
         if (
-            "plot" in self.obj.tags.all()
+            "plot"
+            or "heirloom"
+            or "legacy" in self.obj.tags.all()
             or self.obj.search_tags.all().exists()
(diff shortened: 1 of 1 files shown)


world/crafting/junk_handlers.py
```diff
@@ -39,7 +39,9 @@ def do_junkout(self, caller):
     @property
     def is_plot_related(self):
         if (
-            "plot" in self.obj.tags.all()
+            "plot"
+            or "heirloom"
+            or "legacy" in self.obj.tags.all()
```
TehomCD, line 44: This actually is a subtle bug that won't quite work and would immediately make every item in the game non-junkable. I'll let you try to guess why and tell you if you can't figure it out

TehomCD (changes requested):
Small fix needed

world/crafting/junk_handlers.py
ApostateCD: Yep fixed

TehomCD (approved):

PR #473, opened 2023-07-22
Fix a few bugs with descs - typos and caching

TehomCD:
This one can be summed up with: https://shouldiblamecaching.com/


TehomCD:
Codacy appears to be broken at the moment - not sure if they're down or what

PR #476, opened 2023-08-17
Cast to decimal when receiving mail

TehomCD:
Got a TypeError because decimals and floats can't be added. The `__set__` of the descriptor does a cast, but a `+=` does an addition before the assignment, which is where the error occurs. I'd need to make some wrapper in the `__get__` method that returns some subclass of decimal that allows adding floats by casting them and that's too much to bother with


Tanookami (approved):
Looks great! :shipit:

